PR #1, opened 2016-11-18
Add .stestr directory to .gitignore

masayukig:
We don't need to track this in a git repository.

I'm not sure the ".stestr" is your intentional or not, though :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /build
 /dist
 /.testrepository
+/.stestr
 /testrepository.egg-info
 __pycache__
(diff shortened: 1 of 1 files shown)


mtreinish:
So that actually is intentional, with stestr the state dir is now .stestr, not .testrepository. We might want to change that back  for backwards compat. But this change would have to modify the code writing the state dir.

mtreinish:
Oh, oops I can't read. I don't know why I thought this was a removal not an addition. I guess I needed more coffee

PR #2, opened 2016-12-13
Fix command name - testr to stestr

masayukig:
This commit fixes the command name 'testr' to 'stestr'. I think we
should avoid the command name confliction.

Is this your intention?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ packages =
 
 [entry_points]
 console_scripts =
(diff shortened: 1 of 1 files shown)


mtreinish:
Doesn't this commit conflict with: https://github.com/mtreinish/stestr/commit/6afdd44f54b0ba38f6f4f328742475ecf623492a ?

masayukig:
oops, I was wrong. I was seeing my wrong local branch..

PR #4, opened 2016-12-14
Remove fixtures directory

masayukig:
This commit removes the fixtures directory which only contains
__init__.py now.


mtreinish:
Hah, I just noticed this started passing travis now. But the run command definitely doesn't work yet, it's just returning zero despite not working :)

masayukig:
heh, yeah, I think it's just a starting point :)

PR #5, opened 2016-12-19
Fix concurrency option type to int

masayukig:
This commit fixes the concurrency option from str to int in the run
command. In python3, we need to make it convert to an int value,
explicitly. Otherwise, "TypeError: 'str' object cannot be interpreted as
an integer" error occurs.


mtreinish:
Yeah, it's better to play it safe here. We probably should also fix the type from the argparser option declaration in all the commands, but there's no guarantee a non-int would slip in.

PR #17, opened 2017-01-06
Fix minor subunit2sql things

masayukig:
I just fixed some tiny things.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ __pycache__
 *egg*
 doc/build/
 releasenotes/build/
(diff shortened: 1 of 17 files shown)


mtreinish:
Does github let you make the pull request on my subunit2sql branch. Right now this is going towards master, I'd prefer we add on to the subunit2sql branch, because merging this means we move the whole subunit2sql branch (your patches and mine) onto master, which I think it's still a bit early for

masayukig:
oops, I'll change it on the subunit2sql branch..

PR #25, opened 2017-02-03
Remove extra dash from force-init option of load

masayukig:
This commit removes an extra dash from force-init option of load
subcommand. Three dashes are too many for the option.


mtreinish:
oops, my bad

PR #26, opened 2017-02-10
Add CLI option to stestr run to randomize test order

mtreinish:
This commit adds a new option to randomize the test order. This is
useful because in certain situations you want to introduce a bit
of entropy into the execution order. When this option is used the
same scheduler test partitioning strategy is used, but before the
list of tests is passed to subunit.run for execution the list is
shuffled.

### DIFF
--- a/stestr/tests/test_scheduler.py
+++ b/stestr/tests/test_scheduler.py
@@ -52,6 +52,17 @@ class TestScheduler(base.TestCase):
         self.assertEqual(3, len(partitions[0]))
         self.assertEqual(4, len(partitions[1]))
 
+    def test_random_partitions(self):
+        repo = memory.RepositoryFactory().initialise('memory:')
+        test_ids = frozenset(['a_test', 'b_test', 'c_test', 'd_test'])
+        sorted_parts = scheduler.partition_tests(test_ids, 2, repo, None)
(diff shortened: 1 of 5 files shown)


mtreinish:
@masayukig does this look good to you? I wrote this kinda quickly so I don't have confidence in the patch.

stestr/tests/test_scheduler.py
```diff
@@ -52,6 +52,17 @@ def test_partition_tests(self):
...
 
+    def test_random_partitions(self):
+        repo = memory.RepositoryFactory().initialise('memory:')
+        test_ids = frozenset(['a_test', 'b_test', 'c_test', 'd_test'])
+        sorted_parts = scheduler.partition_tests(test_ids, 2, repo, None)
+        random_parts = scheduler.partition_tests(test_ids, 2, repo, None,
+                                                 randomize=True)
+        self.assertNotEqual(sorted_parts, random_parts,
+                            "The sorted list %s is the same order as the "
+                            "shuffled list %s which is incorrect" % (
```
masayukig: heh, I think this is not always true. Accidentally, it can be same order?

masayukig:
I'm sorry for delayed reply.. I just left one comment about the test.

PR #27, opened 2017-02-15
Fix scheduler.partition_tests randomize option

masayukig:
This commit fixes the scheduler.partition_tests randomize option.
However, we can't test it because the outcome can be same as randomize=False.

### DIFF
--- a/stestr/scheduler.py
+++ b/stestr/scheduler.py
@@ -108,7 +108,7 @@ def partition_tests(test_ids, concurrency, repository, group_callback,
             for partition in partitions:
                 temp_part = list(partition)
                 random.shuffle(temp_part)
-                out_parts.append(set(temp_part))
+                out_parts.append(list(temp_part))
             return out_parts
         else:
             return partitions
--- a/stestr/tests/test_scheduler.py
+++ b/stestr/tests/test_scheduler.py
@@ -55,13 +55,20 @@ class TestScheduler(base.TestCase):
     def test_random_partitions(self):
         repo = memory.RepositoryFactory().initialise('memory:')
         test_ids = frozenset(['a_test', 'b_test', 'c_test', 'd_test'])
-        sorted_parts = scheduler.partition_tests(test_ids, 2, repo, None)
         random_parts = scheduler.partition_tests(test_ids, 2, repo, None,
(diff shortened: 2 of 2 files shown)


stestr/scheduler.py
```diff
@@ -108,7 +108,7 @@ def consume_queue(groups):
             for partition in partitions:
                 temp_part = list(partition)
                 random.shuffle(temp_part)
-                out_parts.append(set(temp_part))
+                out_parts.append(list(temp_part))
```
mtreinish: So why the change in type here? I was casting this with set() because that's what normally gets returned outside the randomize case.

masayukig: oh, I should have mentioned this in the commit message.. This is actually the main reason of this patch. I found that `out_parts.append(set(temp_part))` returns `[{'c_test', 'd_test'}, {'b_test', 'a_test'}]` but `partitions` is `[['c_test', 'd_test'], ['b_test', 'a_test']]` (in python3). So I just fixed it.

```
  File "/Users/igawa/openstack/stestr/stestr/tests/test_scheduler.py", line 64, in test_random_partitions
    sorted_parts, random_parts))
  File "/Users/igawa/openstack/stestr/.tox/py36/lib/python3.6/site-packages/testtools/testcase.py", line 411, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/Users/igawa/openstack/stestr/.tox/py36/lib/python3.6/site-packages/testtools/testcase.py", line 498, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: !=:
reference = [['c_test', 'd_test'], ['b_test', 'a_test']]
actual    = [{'c_test', 'd_test'}, {'b_test', 'a_test'}]
: The sorted list [['c_test', 'd_test'], ['b_test', 'a_test']] is the same order as the shuffled list [{'c_test', 'd_test'}, {'b_test', 'a_test'}] which is incorrect
```

mtreinish: ah, ok. Do you want to rebase this (since github is complaining) and add that to the commit message, then I'll push it through

masayukig: Sure, I'll do that after breakfast :)

PR #28, opened 2017-02-15
Enable until-failure option with no-discover option

masayukig:
This commit enables '--until-failure' option with '--no-discover'
option. I actually wanted this when I tested partition_tests() randomize
option.


mtreinish:
So this looks fine to me, but we are mirroring the logic in _run_tests() at the bottom just with a different source for the Popen object. I wonder if there is some cleanup/de-duplication we can do as a followup here.

PR #29, opened 2017-02-16
Fix scheduler.partition_tests randomize option

masayukig:
(I just updated the commit message with deleting and making the branch. I don't know how to do it without the tricky thing...)

This commit fixes the scheduler.partition_tests randomize option. I
noticed that 'out_parts.append(set(temp_part))' returns
 [{'c_test', 'd_test'}, {'b_test', 'a_test'}]
but 'partitions' is
 [['c_test', 'd_test'], ['b_test', 'a_test']]
(in python3).

However, we can't test it because it can be same as randomize=False in a
small chance. It might be rare, however unstable tests can be confusing.

### DIFF
--- a/stestr/scheduler.py
+++ b/stestr/scheduler.py
@@ -108,7 +108,7 @@ def partition_tests(test_ids, concurrency, repository, group_callback,
             for partition in partitions:
                 temp_part = list(partition)
                 random.shuffle(temp_part)
-                out_parts.append(set(temp_part))
+                out_parts.append(list(temp_part))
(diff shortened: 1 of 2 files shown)


mtreinish:
I think you can always just git push --force to your user branch and github will update things, or at least that's what it has done for me in the past. I do agree the workflow for revisions in github isn't the best and is very confusing.

masayukig:
oh, git push --force, Thank you! I'll try to do it next time :)

PR #31, opened 2017-02-20
Print nice error message when test_path isn't set

sheyman:
If you are running stestr and somehow end up in an
unexpected directory (for example when running within
an IDE) the config file may not be found, and the
test_path variable in stestr/config_file.py will
not be set, resulting in an exception. This patch
will check for a non-None value for test_path and
print a (hopefully) useful error message when
test_path isn't set.

### DIFF
--- a/stestr/config_file.py
+++ b/stestr/config_file.py
@@ -11,6 +11,7 @@
 # under the License.
 
 import re
+import sys
 
 from six.moves import configparser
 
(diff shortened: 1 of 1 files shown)


stestr/config_file.py
```diff
@@ -50,6 +52,11 @@ def get_run_command(self, options, test_ids=None, regexes=None):
             test_path = options.test_path
         elif self.parser.has_option('DEFAULT', 'test_path'):
             test_path = self.parser.get('DEFAULT', 'test_path')
+        else:
+            print("no test_path can be found in either the command line "
```
mtreinish: I would prefer a capital to start the sentence, so s/no/No. But it's not really worth blocking over

PR #34, opened 2017-02-21
fix error where gdbm is not available

sheyman:
Use nbdm rather than gdbm

### DIFF
--- a/stestr/repository/file.py
+++ b/stestr/repository/file.py
@@ -19,7 +19,6 @@ import os
 import sys
 import tempfile
 
-from future.moves import dbm
 from future.moves.dbm import ndbm as my_dbm
 from subunit import TestProtocolClient
 import subunit.v2
--- a/stestr/repository/file.py
+++ b/stestr/repository/file.py
@@ -19,7 +19,7 @@ import os
 import sys
 import tempfile
 
-from future.moves.dbm import ndbm as my_dbm
+from future.moves.dbm import dumb as my_dbm
 from subunit import TestProtocolClient
 import subunit.v2
 import testtools
--- a/stestr/repository/file.py
+++ b/stestr/repository/file.py
@@ -19,7 +19,8 @@ import os
 import sys
 import tempfile
 
-from six.moves import dbm_gnu as dbm
+from future.moves import dbm
+from future.moves.dbm import ndbm as my_dbm
 from subunit import TestProtocolClient
 import subunit.v2
 import testtools
@@ -135,10 +136,10 @@ class Repository(repository.AbstractRepository):
         # May be too slow, but build and iterate.
         # 'c' because an existing repo may be missing a file.
         try:
-            db = dbm.open(self._path('times.dbm'), 'c')
-        except dbm.error:
+            db = my_dbm.open(self._path('times.dbm'), 'c')
+        except my_dbm.error:
             os.remove(self._path('times.dbm'))
-            db = dbm.open(self._path('times.dbm'), 'c')
+            db = my_dbm.open(self._path('times.dbm'), 'c')
         try:
             result = {}
             for test_id in test_ids:
--- a/stestr/repository/file.py
+++ b/stestr/repository/file.py
@@ -258,7 +258,7 @@ class _SafeInserter(object):
         final_path = os.path.join(self._repository.base, str(run_id))
         atomicish_rename(self.fname, final_path)
         # May be too slow, but build and iterate.
-        db = dbm.open(self._repository._path('times.dbm'), 'c')
+        db = my_dbm.open(self._repository._path('times.dbm'), 'c')
         try:
             db_times = {}
             for key, value in self._times.items():
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 # The order of packages is significant, because pip processes them in the order
 # of appearance. Changing the order has an impact on the overall integration
 # process, which may cause wedges in the gate later.
+future
 pbr>=1.8
 python-subunit>=0.18.0
 fixtures>=3.0.0


mtreinish:
I think you doubled up on the commits, shouldn't this just be ndbm patch?

sheyman:
Yep, will fix

masayukig:
In windows, it doesn't work.. It is not related to this patch, though.

```
py27 runtests: commands[0] | stestr run
running=${PYTHON:-python} -m subunit.run discover -t ./ ./stestr/tests
  File "C:\Users\igawa\stestr\.tox\py27\Scripts\stestr-script.py", line 11, in <module>
    load_entry_point('stestr', 'console_scripts', 'stestr')()
  File "c:\users\igawa\stestr\stestr\cli.py", line 103, in main
    sys.exit(args[0].func(args))
  File "c:\users\igawa\stestr\stestr\commands\run.py", line 185, in run
    subunit_out=args.subunit)
  File "c:\users\igawa\stestr\stestr\commands\run.py", line 340, in _run_tests
    return run_tests()
  File "c:\users\igawa\stestr\stestr\commands\run.py", line 327, in run_tests
    proc)) for proc in cmd.run_tests()]
  File "c:\users\igawa\stestr\stestr\test_listing_fixture.py", line 214, in run_tests
    preexec_fn=self._clear_SIGPIPE)
  File "c:\python27\Lib\subprocess.py", line 343, in __init__
    raise ValueError("preexec_fn is not supported on Windows "
ValueError: preexec_fn is not supported on Windows platforms
```

stestr/repository/file.py
```diff
@@ -135,10 +136,10 @@ def _get_test_times(self, test_ids):
         # May be too slow, but build and iterate.
         # 'c' because an existing repo may be missing a file.
         try:
-            db = dbm.open(self._path('times.dbm'), 'c')
-        except dbm.error:
+            db = my_dbm.open(self._path('times.dbm'), 'c')
```
mtreinish: So I'm a bit confused by the distinction here, why do we use dbm.open() when we first create in stopTestRun() and my_dbm here?

sheyman: that was an old merge problem - I will refresh the PR

mtreinish:
It looks better, but it's failing pep8 now: https://travis-ci.org/mtreinish/stestr/jobs/205137380#L183

sheyman:
pep8[happy] = true

masayukig:
This patch works on my macOS environment at least.
```
  System Version:	macOS 10.12.3 (16D32)
  Kernel Version:	Darwin 16.4.0
```

mtreinish:
So I've done some testing and digging into this quagmire of python dbm support. The only version which is guaranteed to be on both python2 and python3 across windows, mac, and linux (and other OSes too) is the dbm dumb module. So I'm thinking we should use that here, the only concern with that is is the potential performance overhead since it's a pure python implementation. I suspect it won't actually matter too much but I guess we should check that before. If the performance is too poor then I'm fine with this as long as we special case it for non-windows systems, and use dbm.dumb for windows (since that's the only thing guaranteed to be there on a windows system)

masayukig:
I'm actually not sure why we need to specify dbm.dumb(py3) or dumbdbm(py2) explicitly... Because in the dbm document[1][2],

> dbm is a generic interface to variants of the DBM database — dbm.gnu or dbm.ndbm. If none of these modules is installed, the slow-but-simple implementation in module dbm.dumb will be used. 

So, I think using `dbm` is the simplest way to use dbm here..? Do I miss something?

[1] https://docs.python.org/3.6/library/dbm.html#module-dbm
[2] https://github.com/python/cpython/blob/3.6/Lib/dbm/__init__.py#L5

masayukig:
And I just tried to get performance difference between mtreinish:master and masayukig:try-dumbdbm https://github.com/mtreinish/stestr/compare/master...masayukig:try-dumbdbm

The result is here. https://dl.dropboxusercontent.com/u/456151/dumb_linux_2.png
 master: mtreinish:master
 dbm: masayukig:try-dumbdbm

So, in this test, the result shows that dbm is faster than master...
The test is just run test_file.py like this. http://paste.openstack.org/show/601712/

mtreinish:
Hmm, interesting, so running with the "slower" pure python implementation is faster at smaller data sets. Let me test it with nova's unit tests to really fill up the dbm file and see if that makes a difference when the dbm file is large. I have a feeling that is where it would make a difference (if any)

@masayukig to answer your question, the reason we have to be explicit here is that the use of anydbm on python2 (which is the equivalent of dbm on python3) will use dbhash in the default case. dbhash was removed in python3 and causes the original bug I was fixing in: https://github.com/mtreinish/stestr/commit/9869bfe48ed87d9abf3f3397c46a9d7f8dc60031 If we were only concerned about python3 I'd agree with you and we'd just use dbm straight on python3.

mtreinish:
ok, I did some cheap tests with nova's unit tests (which is the biggest test suite I interact with at ~15k tests) it matched what you were seeing @masayukig (although your testing didn't cover scheduling which is where the dbm penalty would be if anywhere) There was no real noticeable penalty for using dumb dbm. Here is my data (I didn't bother to graph it but just printed the output from running time) http://paste.openstack.org/show/601795/

I think this is good enough data to just say lets use dumb dbm everywhere and move on. It's guaranteed to work in any environment since it's a pure python implementation. @sheyman can you respin this patch to use dumb dbm instead of ndbm?

sheyman:
changed to dumb dbm

mtreinish:
ok I merged this with: https://github.com/mtreinish/stestr/commit/952d0d2dc8f2dab876caaa26494e536ed0cedc85 I had to do some manually rebasing (and cleaned up the commit message) to make it merge cleanly without polluting the git log. Man, do I hate github, gerrit does all that for you.

PR #39, opened 2017-02-22
Fix error stestr run with bad regex

### LINKED ISSUE #36: An error occurs when `stestr run` with unmatched regex
An error occurs when I run `stestr run` with an unmatched regex like `stestr run foobar`

```
$ stestr run foobar
running=${PYTHON:-python} -m subunit.run discover -t ./ ./stestr/tests --list
arguments:None, None
  File "/Users/igawa/openstack/stestr/.tox/py27/bin/stestr", line 10, in <module>
    sys.exit(main())
  File "/Users/igawa/openstack/stestr/stestr/cli.py", line 102, in main
    sys.exit(args[0].func(args))
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 185, in run
    subunit_out=args.subunit)
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 337, in _run_tests
    return run_tests()
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 334, in run_tests
    repo_url=cmd.options.repo_url)
  File "/Users/igawa/openstack/stestr/stestr/commands/load.py", line 103, in load
    streams = [sys.stdin.buffer]
AttributeError: 'file' object has no attribute 'buffer'
```

And when I use python3.6, an error doesn't occur but it never returns.

masayukig:
This commit fixes an error when `stestr run` with a bad regex.

Fixes #36

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -328,6 +328,9 @@ def _run_tests(cmd, failing, analyze_isolation, isolated, until_failure,
             partial = False
             if (failing or analyze_isolation or isolated):
                 partial = True
+            if not run_procs:
+                print("The specified regex doesn't match with anything.")
+                return 0
             return load.load((None, None), in_streams=run_procs,
                              partial=partial, subunit_out=subunit_out,
                              repo_type=cmd.options.repo_type,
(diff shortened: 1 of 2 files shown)


mtreinish:
I had a different unit test in mind, but we can add that as a followup

Issue #41, opened 2017-02-23
SQL mode: KeyError: 'success' occurs when all tests are failed


masayukig:
In SQL mode, "KeyError: 'success' occurs when all tests are failed.

```
$ stestr -r sql run test_parallel_passing_bad_regex
running=${PYTHON:-python} -m subunit.run discover -t ./ ./stestr/tests --list
running=${PYTHON:-python} -m subunit.run discover -t ./ ./stestr/tests  --load-list /var/folders/q4/1whx01t50xs0n14m502f_6640000gn/T/tmpZMtEl2
Ran 1 tests in 1.832s
FAILED (id=6cece5da-8e89-4a5a-9468-0f35f0c48848, failures=1)
Traceback (most recent call last):
  File "/Users/igawa/openstack/stestr/.tox/py27/bin/stestr", line 10, in <module>
    sys.exit(main())
  File "/Users/igawa/openstack/stestr/stestr/cli.py", line 103, in main
    sys.exit(args[0].func(args))
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 185, in run
    subunit_out=args.subunit)
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 340, in _run_tests
    return run_tests()
  File "/Users/igawa/openstack/stestr/stestr/commands/run.py", line 337, in run_tests
    repo_url=cmd.options.repo_url)
  File "/Users/igawa/openstack/stestr/stestr/commands/load.py", line 141, in load
    case.run(result)
  File "/Users/igawa/openstack/stestr/.tox/py27/lib/python2.7/site-packages/testtools/testsuite.py", line 171, in run
    result.status(**event_dict)
  File "/Users/igawa/openstack/stestr/.tox/py27/lib/python2.7/site-packages/testtools/testresult/real.py", line 467, in status
    _strict_map(methodcaller('status', *args, **kwargs), self.targets)
  File "/Users/igawa/openstack/stestr/.tox/py27/lib/python2.7/site-packages/testtools/testresult/real.py", line 442, in _strict_map
    return list(map(function, *sequences))
  File "/Users/igawa/openstack/stestr/stestr/repository/sql.py", line 289, in status
    self.hook.status(*args, **kwargs)
  File "/Users/igawa/openstack/stestr/.tox/py27/lib/python2.7/site-packages/testtools/testresult/real.py", line 467, in status
    _strict_map(methodcaller('status', *args, **kwargs), self.targets)
[...]

PR #42, opened 2017-02-23
Fix wrong count condition in SQL

masayukig:
This commit fixes the wrong count condition in sql.py. The condition
always count it as success because

  if test_dict['status'] == 'success' or 'xfail':

is always True.

Fixes: #41

### DIFF
--- a/stestr/repository/sql.py
+++ b/stestr/repository/sql.py
@@ -244,9 +244,9 @@ class _SqlInserter(repository.AbstractTestRun):
             else:
                 self.totals[test_dict['status']] += 1
             values = {}
-            if test_dict['status'] == 'success' or 'xfail':
+            if test_dict['status'] in ('success', 'xfail'):
                 values['passes'] = self.totals['success']
-            elif test_dict['status'] == 'fail' or 'uxsuccess':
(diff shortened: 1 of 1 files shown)


stestr/repository/sql.py
```diff
@@ -244,9 +244,9 @@ def _handle_test(self, test_dict):
             else:
                 self.totals[test_dict['status']] += 1
             values = {}
-            if test_dict['status'] == 'success' or 'xfail':
+            if test_dict['status'] in ('success', 'xfail'):
                 values['passes'] = self.totals['success']
-            elif test_dict['status'] == 'fail' or 'uxsuccess':
+            elif test_dict['status'] in ('fail' or 'uxsuccess'):
```
mtreinish: Shouldn't this also be a comment?

mtreinish: err sorry not a comment, a comma

masayukig: ah, yeah, it should be.. (It seems I'm still in jet lag..)

mtreinish:
I kinda wish we had unit tests for the sql backend. But I won't force you to add any

masayukig:
Thanks for merging! heh, yeah, exactly. I was lazy, but I'll do it in another patch.

PR #44, opened 2017-02-27
Add test_sql.py

masayukig:
This commit adds test_sql.py to test sql repository.


masayukig:
hmm, how can we test sql things..? Just do `pip install 'stestr[sql]'` in .travis.yml?

mtreinish:
Yeah I would update the travis.yml to install the extras or maybe add it to test-requirements.txt

PR #46, opened 2017-03-06
Add support to stestr load to append to an existing entry

mtreinish:
This conmmit adds suport to the stestr load command and the underlying
repository API to append a stream into an existing entry in the DB. In
the repository API this is accomplished by adding a new optional
argument on the _get_inserter Abstract repository class to specify a
run_id. When specified the inserter returned will be for that already
existing entry in the repository. The stestr load command just leverages
this new mechanism by taking an id field on the cli and passing it to
the invocation of get_inserter.

### DIFF
--- a/stestr/commands/load.py
+++ b/stestr/commands/load.py
@@ -37,6 +37,9 @@ def set_cli_opts(parser):
     parser.add_argument("--subunit", action="store_true",
                         default=False,
                         help="Display results in subunit format.")
+    parser.add_argument("--id", "-i", default=None,
+                        help="Append the stream into an existing entry in the "
+                             "repository")
 
 
 def get_cli_help():
@@ -124,7 +127,13 @@ def load(arguments, in_streams=None, partial=False, subunit_out=False,
     if args:
         _subunit = getattr(args, 'subunit')
     _subunit_out = _subunit or subunit_out
-    inserter = repo.get_inserter(partial=partial_stream)
+    _run_id = None
+    if args:
+        _run_id = getattr(args, 'id')
+    if not _run_id:
+        inserter = repo.get_inserter(partial=partial_stream)
+    else:
+        inserter = repo.get_inserter(partial=partial_stream, run_id=_run_id)
     if _subunit_out:
         output_result, summary_result = output.make_result(inserter.get_id)
     else:
--- a/stestr/repository/abstract.py
+++ b/stestr/repository/abstract.py
@@ -67,7 +67,7 @@ class AbstractRepository(object):
         """
         raise NotImplementedError(self.get_failing)
 
-    def get_inserter(self, partial=False):
+    def get_inserter(self, partial=False, run_id=None):
         """Get an inserter that will insert a test run into the repository.
 
         Repository implementations should implement _get_inserter.
@@ -81,9 +81,9 @@ class AbstractRepository(object):
             that testtools 0.9.2 and above offer. The startTestRun and
             stopTestRun methods in particular must be called.
         """
-        return self._get_inserter(partial)
+        return self._get_inserter(partial, run_id)
 
-    def _get_inserter(self):
+    def _get_inserter(self, partial=False, run_id=None):
         """Get an inserter for get_inserter.
 
         The result is decorated with an AutoTimingTestResultDecorator.
--- a/stestr/repository/file.py
+++ b/stestr/repository/file.py
@@ -128,8 +128,8 @@ class Repository(repository.AbstractRepository):
                 raise
         return _DiskRun(run_id, run_subunit_content)
 
-    def _get_inserter(self, partial):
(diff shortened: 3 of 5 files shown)


mtreinish:
@masayukig any thoughts on this?

The repository code here is a bit hairy since the way it works is to treat a test run in the repository as a subunit stream / testtools stream but instead of returning the result it writes it to the repository. I wish it was easier to follow, there is definitely some cleanup that can happen here. But, that will have to be in a separate patch.

mtreinish:
Pushed an update to rebase it on master. I think this is good to go now

stestr/commands/load.py
```diff
@@ -124,7 +127,13 @@ def make_tests():
...
         _subunit = getattr(args, 'subunit')
     _subunit_out = _subunit or subunit_out
-    inserter = repo.get_inserter(partial=partial_stream)
+    _run_id = None
+    if args:
+        _run_id = getattr(args, 'id')
+    if not _run_id:
+        inserter = repo.get_inserter(partial=partial_stream)
+    else:
+        inserter = repo.get_appender(partial=partial_stream, run_id=_run_id)
```
masayukig: Where's this method definition?

stestr/repository/abstract.py
```diff
@@ -81,7 +81,7 @@ def get_inserter(self, partial=False):
             that testtools 0.9.2 and above offer. The startTestRun and
             stopTestRun methods in particular must be called.
         """
-        return self._get_inserter(partial)
+        return self._get_inserter(partial, run_id)
 
     def _get_inserter(self):
```
masayukig: Do you want to make this `_get_inserter(self, run_id=None)` ?

mtreinish:
Oops, I guess I was very sloppy and didn't check my work well enough. Those are leftovers from an earlier attempt when I used a separate method. (but it turns out I didn't need it) I'll fix that in the next revision.

As for the use case this is the first part in implementing the --combine flag for stestr run for real. We need to be able to load streams into an existing repository entry to do that. I would have pushed that up as a branch in gerrit, but it needs to be a separate review and github doesn't let you do that well.

stestr/repository/abstract.py
mtreinish: So this is a weird case it's a method that doesn't really have a definition. I would have partial too, since that's missing as well. I can do that but I figured just leave it as is. But now that I think about it's kinda a weird to have an abstract class that doesn't define the actual interface for the method. I'll update this.

stestr/commands/load.py
mtreinish: It was a leftover from a earlier iteration. I'll fix it, good catch

mtreinish:
Ok, let's see if I got it right this time

masayukig:
Thanks, this works for me. But I've noticed that a specified id's result was removed. I'm not sure this is intended or not. I'm just curious.

For example:
```
$ ls -l .stestr
total 100
-rw------- 1 igawa staff 23290  3  9 18:40 0
-rw------- 1 igawa staff 23290  3  9 18:41 1
-rw------- 1 igawa staff     0  3  9 18:41 failing
-rw-r--r-- 1 igawa staff     2  3  9 18:40 format
-rw-r--r-- 1 igawa staff     2  3  9 18:41 next-stream
-rw-r--r-- 1 igawa staff  4578  3  9 18:41 times.dbm.bak
-rw-r--r-- 1 igawa staff 26120  3  9 18:41 times.dbm.dat
-rw-r--r-- 1 igawa staff  4578  3  9 18:41 times.dbm.dir
$ cat .stestr/0 | subunit-1to2| stestr load -i 1
Ran 52 tests in 19.682s (+7.468s)
PASSED (id=2)
$ ls -l .stestr
total 124
-rw------- 1 igawa staff 23290  3  9 18:40 0
-rw------- 1 igawa staff 47321  3  9 18:42 2
-rw------- 1 igawa staff     0  3  9 18:42 failing
-rw-r--r-- 1 igawa staff     2  3  9 18:40 format
-rw-r--r-- 1 igawa staff     2  3  9 18:42 next-stream
-rw-r--r-- 1 igawa staff  4578  3  9 18:42 times.dbm.bak
-rw-r--r-- 1 igawa staff 26120  3  9 18:42 times.dbm.dat
-rw-r--r-- 1 igawa staff  4578  3  9 18:42 times.dbm.dir
```

mtreinish:
That wasn't my intent, but it doesn't really break anything. I probably should fix it though, it should be pretty easy.

mtreinish:
Ok, it should be writing into the existing stream in the repository now and not deleting the old one/creating a new file. Heh, hopefully I've got everything working now.

mtreinish:
@masayukig any thoughts on the latest revision?

masayukig:
Thank you! It works for me :)

PR #49, opened 2017-04-06
Enable coverage with tox

masayukig:
This commit enables to get coverage with tox. We probably should care
about unit test coverage data of stestr a bit :)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,6 +22,11 @@ commands =
 [testenv:venv]
 commands = {posargs}
 
+[testenv:cover]
+commands =
+    coverage run stestr/cli.py run {posargs}
+    coverage html -d cover
+
 [testenv:docs]
 commands = python setup.py build_sphinx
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -22,6 +22,11 @@ commands =
 [testenv:venv]
 commands = {posargs}
 
+[testenv:cover]
+commands =
+    coverage run setup.py test {posargs}
```
mtreinish: So my only issue with this is it's running testr and not stestr. I think it would be ideal if we used stestr to run the testing for coverage too.

tox.ini
```diff
@@ -22,6 +22,11 @@ commands =
 [testenv:venv]
 commands = {posargs}
 
+[testenv:cover]
+commands =
+    coverage run stestr/cli.py run {posargs}
```
masayukig: How about like this?

mtreinish: Yep, that should be fine.

PR #50, opened 2017-04-07
Improve coverage collection

masayukig:
This commit improves to collect more coverage data. According to the
coverage.py document[1], we need some tricks to measure sub-processes
coverage. So, this commit is trying to do it.

[1] http://coverage.readthedocs.io/en/coverage-4.2/subprocess.html

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,3 +2,4 @@
 branch = True
 source = stestr
 omit = stestr/tests/*
+parallel = True
--- a/tox.ini
+++ b/tox.ini
@@ -23,8 +23,12 @@ commands =
 commands = {posargs}
 
 [testenv:cover]
+setenv =
+    VIRTUAL_ENV={envdir}
+    PYTHON=coverage run --source stestr
 commands =
     coverage run stestr/cli.py run {posargs}
+    coverage combine
     coverage html -d cover
 
 [testenv:docs]


masayukig:
This fix is not perfect, actually. The coverage rate of stestr/repository/sql.py is still low against actual executing. But it's better. We should dig it more, though.

mtreinish:
I think you're on the right track, but what I think is the proper way to handle this is:

http://paste.openstack.org/show/605755/

Instead of trying to trick the python interpreter to launch with coverage enabled (which is essentially what the doc you linked, and I sent you last night does) we explicitly tell stestr to launch coverage run as the python interpreter in subprocess. All of the subprocess commands respect the PYTHON env var:

https://github.com/mtreinish/stestr/blob/master/stestr/config_file.py#L66 (except on win32, but that doesn't really matter for coverage) 

If we launch the launch the subprocesses under coverage and set things up to combine reports this should get us full coverage. When I ran this locally I got a 58% coverage report which I think covered everything we are actually running.

masayukig:
Oh, Thanks. It's really better than mine. I just updated my patch with your suggestion.

masayukig:
oh, 63.803%! :)

stestr/tests/base.py
```diff
@@ -18,6 +19,7 @@ class TestCase(testtools.TestCase):
 
     def setUp(self):
         super(TestCase, self).setUp()
+        coverage.process_startup()
```
mtreinish: Is this still needed?

masayukig: nope.. :( I'll remove it soon

PR #53, opened 2017-04-08
Add codeclimate and pypi version badges

masayukig:
This commit adds two badges of codeclimate[1] and pypi version[2]. This
might be useful for healthy code and to know the latest version of
stestr.

[1] https://codeclimate.com/github/mtreinish/stestr
[2] https://pypi.python.org/pypi/stestr

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,6 +7,8 @@ Slim/Super Test Repository
 .. image:: https://coveralls.io/repos/github/mtreinish/stestr/badge.svg?branch=master
     :target: https://coveralls.io/github/mtreinish/stestr?branch=master
 
+.. image:: https://img.shields.io/pypi/v/stestr.svg
+    :target: https://pypi.python.org/pypi/stestr
 
 You can see the full rendered docs at: http://stestr.readthedocs.io/en/latest/
 


mtreinish:
Do you find the code climate useful? I'm looking at it and the only thing it seems to tell me is there are similar chunks somewhere else and cyclomatic complexity is too high. Neither is really actionable and can be misleading metrics.

mtreinish:
Oh, and todos in the code :)

masayukig:
heh, I feel this is not so useful so far. Because all of the issues severity are minor. But I think the trend graph[1] might be useful. If the trend is getting worse, it could be something wrong and it might be a good timing for refactoring.

But I'm not stick on this. So I can drop the codeclimate badge if you don't like it :)

[1] https://codeclimate.com/github/mtreinish/stestr/trends

mtreinish:
Yeah, I'm thinking lets keep tracking it, if it does end up providing us useful data we can add the badge then. But for right now lets stick with just adding the pypi badge

masayukig:
OK, I'll drop the codeclimate badge :)

PR #56, opened 2017-04-27
Cleanup error handling for missing test-path or config file

mtreinish:
This commit cleans up the error handling for when stestr run is called
without a repository and no config file or test-path arg is specified.
This is very common for new users to go to a directory with tests and
run 'stestr run' not having read the README or manual. Previously if
this was done you'd be greated by a repository not found traceback which
is hardly helpful. This cleans it up so a message prints explaing what's
missing and exits with an error code. As commit also cleans up the
wording for the error message when the same mistake is made with a
pre-existing repository.


mtreinish:
@masayukig didn't I give you merge access? If not I can try to fix that.

PR #57, opened 2017-04-27
Cleanup output from test listing fixture

mtreinish:
The test listing fixture had some leftover output straight from testr
which can be confusing. First is it prints the subprocess commands that
were being run, which is unecessary and confusing if you don't know what
is going in on internally. Secondly when it encounters an import it
raises an unhandled exception to crash. This commit fixes this by
printing out a more detailed view of what happened with an explanation.
Also a leftover failure condition from testrepository was removed
because it can never be tripped in stestr.

### DIFF
--- a/stestr/output.py
+++ b/stestr/output.py
@@ -93,20 +93,6 @@ def make_result(get_id, output=sys.stdout):
     return result, summary
 
 
-def output_values(values, output=sys.stdout):
-    """Display key value pairs.
-
-    :param values: A list of tuples for key value pairs. Each tuple in the list
-        is of the form (key, value).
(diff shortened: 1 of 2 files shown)


mtreinish:
man, coveralls likes to spam...

masayukig:
heh, just ignore or filter out notifications from coveralls.
Or try to use the other service like https://codecov.io/ ?

masayukig:
> Yeah, that's why I did it. I miss the days when you had a table of exit codes and a reasons why :). But, I can change it to be 1 or any other number if you prefer.

I'm OK with that code 100. I don't have particular preference with this exit code so far :)

PR #58, opened 2017-04-27
Update requirements

masayukig:
This commit updates requirements. We should keep up-to-date the python
libraries as possible.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,9 +2,9 @@
 # of appearance. Changing the order has an impact on the overall integration
 # process, which may cause wedges in the gate later.
 future
-pbr>=1.8
-python-subunit>=0.18.0
-fixtures>=3.0.0
-six>=1.10.0
-testtools>=2.2.0
-PyYAML>=3.1.0
+pbr!=2.1.0,>=2.0.0 # Apache-2.0
+python-subunit>=0.18.0 # Apache-2.0/BSD
+fixtures>=3.0.0 # Apache-2.0/BSD
+six>=1.10.0 # MIT
+testtools>=2.2.0 # MIT
+PyYAML>=3.10.0 # MIT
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -2,9 +2,9 @@
...
-pbr>=1.8
-python-subunit>=0.18.0
-fixtures>=3.0.0
-six>=1.10.0
-testtools>=2.2.0
-PyYAML>=3.1.0
+pbr!=2.1.0,>=2.0.0 # Apache-2.0
+python-subunit>=0.0.18 # Apache-2.0/BSD
+fixtures>=3.0.0 # Apache-2.0/BSD
+six>=1.9.0 # MIT
```
mtreinish: Isn't this less than what was in there before?

requirements.txt
```diff
@@ -2,9 +2,9 @@
...
+testtools>=1.4.0 # MIT
```
mtreinish: Same with this one, testtools 1.4.0 < 2.2.0

mtreinish:
So I get what you did here, but some of the things in stestr were already requiring newer versions than what's in OpenStack's global requirements.

masayukig:
ah, yeah, I should combine them not only use the OpenStack requirements.
 will fix it

test-requirements.txt
```diff
@@ -3,8 +3,8 @@
...
 
 hacking<0.12,>=0.11.0 # Apache-2.0
-sphinx!=1.3b1,<1.4,>=1.2.1 # BSD
+sphinx>=1.5.1 # BSD
 mock>=2.0 # BSD
-coverage>=3.6 # Apache-2.0
-reno>=2.0.3 # Apache-2.0
 subunit2sql>=1.8.0
+coverage>=4.0 # Apache-2.0
+reno>=1.8.0 # Apache-2.0
```
mtreinish: Here's another openstack global-requirements leak, but it's not a big deal. We really don't use reno for real yet so it's not like we really need it >=2.0.3

PR #64, opened 2017-06-29
Add a concurrency option to the manual scheduler

mtreinish:
This commit adds a new option in the manual scheduler yaml syntax. There
are certain situations where you want to treat a subset of tests being
manually scheduled like the default case with the standard test
scheduling mechanisms. To enable doing this a 'concurrency' field can be
used on a worker, and tests on those workers will be scheduled like
normal across the number of workers specified.

### DIFF
--- a/stestr/test_listing_fixture.py
+++ b/stestr/test_listing_fixture.py
@@ -227,7 +227,8 @@ class TestListingFixture(fixtures.Fixture):
             if hasattr(self.options, 'randomize'):
                 randomize = self.options.randomize
             test_id_groups = scheduler.generate_worker_partitions(
-                test_ids, self.worker_path, randomize)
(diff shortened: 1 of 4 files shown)


stestr/test_listing_fixture.py
```diff
@@ -227,7 +227,8 @@ def run_tests(self):
             if hasattr(self.options, 'randomize'):
                 randomize = self.options.randomize
             test_id_groups = scheduler.generate_worker_partitions(
-                test_ids, self.worker_path, randomize)
+                test_ids, self.worker_path, self.repository,
+                self.group_callback, randomize)
```
masayukig: s/self.group_callback/self._group_callback

mtreinish: Done

Issue #21, opened 2017-01-06
test_listing_fixture is a bad name


mtreinish:
The name for the module stestr.test_listing_fixture and it's class name TestListingFixture are a terrible name and misleading in their function. It sounds like it runs tests on the listing_fixture. The problem is I'm not exactly sure how to describe it.

Currently the class TestListingFixture defines a [fixture](https://pypi.python.org/pypi/fixtures) for the lifecycle of a temp file that contains a list of test_ids. These test_ids are either provided as an input parameter or found through running test discovery (by launching subunit.run discover --list). The fixture also provides functions that require the existence of that temporary test list, which includes the launching of the test runner workers. This means that the test_listing_fixture is where we actually run the test processes. (which is the source of the confusion around the name)

If anyone has an idea for a better name feel free to push a patch renaming things, or just comment on the bug.

shriya999:
how about test_processor_fixture as the main action is processing a test according to a list of test ids?

mtreinish:
Hmm, I like that direction. Maybe we make the module name test_processor.py and the class TestProcessorFixture. I think that will work well. Would you like to submit a pull request making that switch? If not I can do it, but since you came up with the name you should get the commit in the log. :)

shriya999:
Yes i would like to make a pull request as this will be my first bug!! Thanks for reviewing it so fast 🗡

mtreinish:
Sorry, I missed your comments here. I believe github has a method for creating git commits via their web UI. Although, I've never personally used it, because the normal git workflow is much faster IMHO. So you create a pull request you will need to have git on your local machine and create a git commit

Unfortunately I don't really think waiting until May is an option.

mtreinish:
Sorry, I missed your comments here. I believe github has a method for creating git commits via their web UI. Although, I've never personally used it, because the normal git workflow is much faster IMHO. So you create a pull request you will need to have git on your local machine and create a git branch with a commit with the renamed files (and import statements that were using the old name). Once that's done you push the branch up to your github fork and create a pull request from the fork.

Here are some docs I found on how to do this:

https://guides.github.com/introduction/flow/

https://help.github.com/articles/about-pull-requests/

http://rogerdudler.github.io/git-guide/

This one includes links about doing everything in github's web UI:

https://help.github.com/articles/github-flow/

Unfortunately I don't really think waiting until May is an option. I want to include this change in the next release so we can start to stabilize the Python API; which this is the first step towards.

[...]

PR #65, opened 2017-07-20
Rename the test_listing_fixture to test_processor

mtreinish:
The name test_listing_fixture was a bit confusing. It was a hold over
from test_repository that was never renamed. Even in testrepository the
function of the module is much more than simply listing tests. It's
where all the test workers are launched. While if you look at the
internals of the module the life of the fixture is around the temporary
test list files it creates. But, this name is confusing for people not
super familiar with the internals.

To try and clear up any potential confusion here this commit renames it
to the more generic 'test_processor' (and the Fixture class inside).
This hopefully makes it more clear that the role of the module is around
processing the tests, and more than just listing things.

Closes Issue #21

### DIFF
--- a/stestr/config_file.py
+++ b/stestr/config_file.py
@@ -16,7 +16,7 @@ import sys
 from six.moves import configparser
 
 from stestr.repository import util
-from stestr import test_listing_fixture
+from stestr import test_processor
 
 
 class TestrConf(object):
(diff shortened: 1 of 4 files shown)


mtreinish:
This is going to conflict with my other open pull request https://github.com/mtreinish/stestr/pull/66 which reworks the entire internal python interface. We should merge that much larger change before this one. It'll be much easier to rebase this.

stestr/config_file.py
```diff
@@ -33,19 +33,19 @@ def __init__(self, config_file):
         self.config_file = config_file
 
     def get_run_command(self, options, test_ids=None, regexes=None):
-        """Get a test_listing_fixture.TestListingFixture for this config file
+        """Get a test_processor/.TestProcessorFixture for this config file
```
mtreinish: Nope that was a mistake, I'll fix it in a new revision. I'm going to wait until after #66 merges before I bother though.

masayukig: OK :) I'll look at #66 , anyways.

mtreinish:
I rebased this on current head, which includes #66 now, and also fixed the typo you pointed out

masayukig:
thanks! lgtm

Issue #8, opened 2016-12-31
CLI arguments should use a standarized data structure and interface


mtreinish:
Currently the stestr.cli module passes the raw output from argparse's parse_known_args() method to the run() method in each command module. While this works it's not a very friendly interface and it makes using the commands via a python interface a bit more complicated since you'd have to construct a tuple with the argparse Namespace object and any additional parameters to pass arguments into the run() methods.

To really make the commands externally consumable via a python api we need to standardize how we pass arguments into these functions. Whether that consists of switching from the tuple with the Namespace and list to something like a dict, or leaving it as is doesn't really matter too much. We just need a clearly writtend doc defining how parameters are passed in and having examples on how to use it outside of stestr.cli.

mtreinish:
The idea I had in mind right now is for the commands run() function have that call a function that generates a couple of dicts with a known schema (maybe using voluptuous to define it, but that's probably overkill) that can be used by an inner function that can be called with at least one of those dicts used as **kwargs directly. That way for python consumers they can just call the inner functions from outside of stestr that will have a defined set of kwargs they can use. Then for the interface docs we document how to call those inner functions.

Where this gets a bit complicated is when you look at the call stack and how variables are accessing the arguments object. Right now it's kinda used a dumping ground for local state. (it was much easier to just put everything in there and pass the one object around) So things which aren't checked inside the commands/*.py module get used 2 or 3 calls down the stack. So we have to make sure that all those bits are clearly exposed at the top level and at each layer going down the stack.

mtreinish:
I've started working on this idea in my wip branch: https://github.com/mtreinish/stestr/tree/add_python_interface it's still far from complete, but it's the starting point. (also it's untested at this point as well) I'll continue to update that branch as I work on it. I'll PR it when I think it's ready. If someone would like to help on that feel free to just ask, I'm not sure how fast I'll be able to move on it.

mtreinish:
With #66 merged I think this bug is now. I still need to add documentation on how to use the interfaces but I will do that in later patches

PR #66, opened 2017-07-24
Add python interface for commands

mtreinish:
This commit adds a real api for running the stestr commands. Instead of just having a magic run() function that takes in an tuple of a argparse Namespace and a list of undefined arguments this migrates all the real work into a function that has properly defined kwargs. The run() command is then relegated to just convert the Namespace into a dict and pass the arguments into that real function. This enables external programs to just call the new functions with defined args and run commands exactly like on the cli, but with a defined python interface. It makes everything a lot easier for python consumption. The tradeoff here though is that everything is bit more verbose, but that's the cost of being explicit with a defined interface.

As a side effect of this change instead of passing that Namespace object around between all the lower layers real interfaces have to be defined for all the functions. This means a ton of new kwargs, but again this is better in the long run because it means we have defined interfaces for all the functions.

Closes Issue #8

### DIFF
--- a/stestr/commands/failing.py
+++ b/stestr/commands/failing.py
@@ -59,13 +59,35 @@ def _make_result(repo, list_tests=False):
 
 def run(arguments):
     args = arguments[0]
-    repo = util.get_repo_open(args.repo_type, args.repo_url)
+    return failing(repo_type=args.repo_type, repo_url=args.repo_url,
+                   list_tests=args.list, subunit=args.subunit)
+
+
+def failing(repo_type='file', repo_url=None, list_tests=False, subunit=False):
+    """Return the failing tests from the most recent run in the repository
+
+    Note this function depends on the cwd for the repository if `repo_type` is
+    set to file and `repo_url` is not specified it will use the repository
+    located at CWD/.stestr
+
+    :param str repo_type: This is the type of repository to use. Valid choices
+        are 'file' and 'sql'.
+    :param str repo_url: The url of the repository to use.
+    :param bool list_test: Show only a list of failing tests.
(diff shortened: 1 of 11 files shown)


mtreinish:
This is going to conflict with my other open pull request https://github.com/mtreinish/stestr/pull/65 which renames the test_listing_fixture module. We should merge this much larger change before the other. It'll be much easier to rebase the rename.

mtreinish:
Hmm, that wasn't an intentional change. I didn't even notice that. My guess is that I'm returning somewhere I wasn't before and it's passing that out. Well that or there is a debug print I left in somewhere.

mtreinish:
Yep, it was the repo initialization call, that was returning a repository object. Which was being passed to stdout when run. I removed those returns and it fixes the weird output.

mtreinish:
I hate github, this branch merging strategy makes no sense

mtreinish:
ok, finally got it to act correctly, it required me to hand pick the changes.

masayukig:
Thanks, LGTM. This works in my environment at least.

mtreinish:
ok cool, I'm going to merge this then. I've got a bunch of other work behind this change. We can always fix anything I slipped up on after the fact. I'm not planning on pushing a release at least for a little while

afrittoli:
I started reviewing this a bit late, sorry about that.
I did a bit of testing and checking of the new interface and it looks really nice.

It may be handy to have an __init__ file that groups all available APIs in a single module, so that different API do not have to be imported from separate modules. Usage example:

```
from stestr import api

api.list(repo_type='file')
api.run(...)
```

Issue #68, opened 2017-07-25
Running with --subunit flag always returns 0


mtreinish:
If you run stestr run with the --subunit flag the return code is always 0, even if tests fail. stestr run should return a 1 on failed tests. This causes the--until-failure flag to loop forever if running with --subunit. This is a bug carried over from testrepository in: https://bugs.launchpad.net/testrepository/+bug/1411804

mtreinish:
With #69 merged this bug is closed

PR #69, opened 2017-07-26
Fix stestr run --until-failure with subunit output

mtreinish:
The --subunit flag on stestr run, load, etc always returns 0. This is
done primarily to ensure that subunit processing can happen without
issue, especially from inside a python caller. (there is also an implicit
assumption that if you emit subunit whatever consumes that will check
for the failure condition and handle that itself) This has a side effect
for the --until-failure flag on stestr run which will never exit if
--subunit is used because there is never a failure condition detected
because subunit output always returns 0. This commit fixes this by
checking the repository directly for failure if --subunit and
--until-failure are used together.

Closes Issue #68

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -17,6 +17,7 @@ import os
 import subprocess
 
 import six
+import subunit
 import testtools
 
 from stestr.commands import load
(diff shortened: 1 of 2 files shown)


stestr/commands/run.py
```diff
@@ -206,6 +207,21 @@ def run_tests():
             result = run_tests()
```
masayukig: If this first time run_tests() fails with using subunit, we miss the failure. Do I miss something?

mtreinish: oh, nope you're right. In my tests I was was checking that it fails eventually. Looking closely I see there are 2 failures now. So that's definitely what's happening. I'll respin this tomorrow to fix the bug

mtreinish: Ok, this should be fixed now in the latest revision

PR #70, opened 2017-07-26
Remove the stats command

mtreinish:
The stats command didn't really do anything useful, in practice it just
printed the number of runs in the database. When it was ported from
testrepository the vestigal comment was left in too. Since there is no
real use for the command this commit just removes it.


masayukig:
heh, LGTM. I was thinking this was a real stats command like showing total number of execution, failures and successful, and execution time, etc. But this is not the one so far.
If we need it, we can implement it later.

Issue #22, opened 2017-01-19
Improve output formatting


mtreinish:
The subunit-trace module from os-testr provides a much more informative output filter for tests. (see: http://git.openstack.org/cgit/openstack/os-testr/tree/os_testr/subunit_trace.py ) We should port this to stestr and make it the default output. We can keep the current summary output mode as an option, but having the details from subunit-trace will be useful.

mtreinish:
I started working on this here: https://github.com/mtreinish/stestr/tree/use-subunit-trace it still needs a bit of work and I'm not exactly happy with the amount of duplicated code there. But at least for a start it shows how to accomplish the goal of this issue.

mtreinish:
With #72 merged this is now fixed.

PR #72, opened 2017-07-27
Use subunit-trace

mtreinish:
This adds subunit-trace output filtering to stestr. The code is copied directly from the os-testr project [1]. This enables subunit-trace by default on the run and last commands, and enables it as an option on the load command. Right now most of the subunit-trace output options are hard coded, with the only exception being colorized output which there is a config flag for. In the future we might add support for more customized output.

One thing to note is that to avoid a potential namespace collision this does not add the console_script entrypoint for the subunit-trace command. If/when os-testr switches to using stestr we might be able to figure out a deprecation path in os-testr and remove the forked copy of subunit-trace.

Closes Issue #22 

[1] http://git.openstack.org/cgit/openstack/os-testr


stestr/commands/load.py
```diff
@@ -116,6 +129,16 @@ def make_tests():
...
+    if pretty_out:
+        outcomes = testtools.StreamToDict(
+            functools.partial(subunit_trace.show_outcome, sys.stdout,
+                              enable_color=color))
+        summary_result = testtools.StreamSummary()
+        output_result = testtools.CopyStreamResult([outcomes, summary_result])
+        output_result = testtools.StreamResultRouter(output_result)
+        cat = subunit.test_results.CatFiles(sys.stdout)
+        output_result.add_rule(cat, 'test_id', test_id=None)
+        start_time = datetime.datetime.utcnow()
```
masayukig: This causes an error[1] when pretty_out is True(--no-subunit-trace specified).
[1] ```UnboundLocalError: local variable 'start_time' referenced before assignment```

mtreinish: Yep, start time isn't defined in the case pretty_out is false. So when L157 is run there is no start time defined. I should have put start time definition out one indent level and to L144. I'll respin it right now

mtreinish:
ok, the latest revision should have fixed the issue you caught. Thanks for testing and reviewing this and all the other patches this week.

masayukig:
(I really don't know the github workflow... :-p)

PR #75, opened 2017-07-28
Allow to import commands APIs from a single module

afrittoli:
The APIs available to run various stestr commands are distributed
in multiple modules. For ease of use make them all accessible from
the commands module.


stestr/commands/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+from stestr.commands.failing import failing
+from stestr.commands.init import init
+from stestr.commands.last import last
+from stestr.commands.list import list_command
+from stestr.commands.load import load
```
mtreinish: So this has a name conflict because we're overriding the module definition with the function. So when stestr.commands.run goes to call stestr.commands.load.load() things blow up on an attribute error because the function doesn't have a load() attribute.

afrittoli: Heh I see. I updated the change using aliases but I'm not sure that's enough. I should probably add a unit tests about using these?

mtreinish:
I was actually thinking the inverse case, change the function name, but this works too. Yeah a unit test would be nice, although I'm not sure what you'll be testing

mtreinish:
The only open question is whether you think we should update the API docs about this interface instead of importing directly from the per command modules?

mtreinish:
I decided to just merge this as is, if you can please push a doc update patch. It'll be good to land a doc update before we push the next release.

PR #77, opened 2017-08-01
Api module docs

afrittoli:
Three commits:
- fix the docs job fist by adding the subunit_trace module into the tree
- add docs env to travis
- fix the __all__ definition in __init__, it should use strings - and add it to the docs along with an example


mtreinish:
So I'm fine with this, but I'm a bit confused because the docs seem to be rendering fine:

https://readthedocs.org/projects/stestr/builds/

PR #78, opened 2017-08-02
Handle stestr init with existing path

mtreinish:
Previously if you ran stestr init with a file type repo on a path where
the repository already created you would get a stack trace. That's
because the file layer will go to do a mkdir and if the path already
exists it will raise an exception which bubbles up to the cli layer.
This commit just catches the FileExistsError and prints a message
instead.

### DIFF
--- a/stestr/commands/init.py
+++ b/stestr/commands/init.py
@@ -12,6 +12,8 @@
 
 """Initialise a new repository."""
 
+import sys
+
 from stestr.repository import util
 
 
@@ -20,7 +22,7 @@ def run(arguments):
     init(args.repo_type, args.repo_url)
 
 
-def init(repo_type='file', repo_url=None):
+def init(repo_type='file', repo_url=None, stdout=sys.stdout):
     """Initialize a new repository
 
     This function will create initialize a new repostiory if one does not
@@ -38,8 +40,16 @@ def init(repo_type='file', repo_url=None):
         for failures.
     :rtype: int
     """
-
-    util.get_repo_initialise(repo_type, repo_url)
+    try:
+        util.get_repo_initialise(repo_type, repo_url)
+    except OSError as e:
+        if e.errno != 17:
+            raise
+        repo_path = repo_url or './stestr'
+        stdout.write('The specified repository directory %s already exists. '
+                     'Please check if the repository already exists or '
+                     'select a different path\n' % repo_path)
+        return 1
 
 
 def set_cli_opts(parser):


stestr/commands/init.py
```diff
@@ -38,8 +40,11 @@ def init(repo_type='file', repo_url=None):
         for failures.
     :rtype: int
     """
-
-    util.get_repo_initialise(repo_type, repo_url)
+    try:
+        util.get_repo_initialise(repo_type, repo_url)
+    except FileExistsError:
+        stdout.write('The specified repository already exists\n')
+        return 0
```
mtreinish: I'm not sure about this return code. I kinda want to make this a failure condition, because we can't actually be sure if it's a stestr file repository or not from this. Just that the directory exists. On the sql side it's an idempotent function, but that's because it has a concrete api endpoint and the migration scripts it runs are idempotent. But because it's idempotent on sql I thought I should return 0 here, but there isn't actually a guarantee the repository exists in this case. 

Does anyone have any thoughts on this? I'm strongly leaning towards changing this to be a failure condition before we merge this.

masayukig: yeah, I was thinking +1 for a failure condition here originally. But I checked some commands behavior.

* git: success (git init && git init; echo $?)
* hg: failure (hg init && hg init; echo $?)
* mkdir: success (mkdir foo && mkdir foo; echo $?)
* tempest: failure (tempest init foo && tempest init foo; echo $?)

And I found that git-init document[1], it says,
```
Running git init in an existing repository is safe. It will not overwrite things that are already there. The primary reason for rerunning git init is to pick up newly added templates (or to move the repository to another place if --separate-git-dir is given).
```
I think this is a user-friendly interface. So, my conclusion is we can return 0 here :) 

[1] https://git-scm.com/docs/git-init

mtreinish: Well except in that example git actually checks and updates the .git dir. The difference here is that we just know the directory already exists not that it's a stestr repository. Which I think is exactly what tempest and hg are doing. Also on my system mkdir returns 1 if you try to create a dir on an existing path. (coreutils 8.27)

masayukig: oh, my bad about mkdir, my 'mkdir' was aliased to 'mkdir -p'... :-p So, it returns 1 when I unaliase it.

I understand what your concern now. Hmm, I'm fine ```return 1``` here. I think it should be warning. And I think the message can be improved like 'The specified repository already exists. You should check the repository and remove it if you really want to initialize.\n' or something like that.

PR #79, opened 2017-08-08
Add unit tests for return code from functions

mtreinish:
This commit adds unit tests calling the command functions from the
python interface to verify their return codes. As things start to
consume the python interfaces it will be good to point to tests that
validate the return codes and functionality are the same as running from
the CLI. The tests added here are just a mirror of our cli return code
tests but instead of using subprocess they call the python api.

### DIFF
--- a/stestr/tests/test_return_codes.py
+++ b/stestr/tests/test_return_codes.py
@@ -20,11 +20,14 @@ import shutil
 import subprocess
 import tempfile
 
+import fixtures
 import six
 from six import StringIO
 import subunit as subunit_lib
 import testtools
 
+from stestr.commands import list as list_cmd
+from stestr.commands import run
 from stestr.tests import base
 
 
@@ -56,6 +59,24 @@ class TestReturnCodes(base.TestCase):
         self.addCleanup(os.chdir, os.path.abspath(os.curdir))
         os.chdir(self.directory)
         subprocess.call('stestr init', shell=True)
+
+    def _check_subunit(self, output_stream):
+        stream = subunit_lib.ByteStreamToStreamResult(output_stream)
+        starts = testtools.StreamResult()
+        summary = testtools.StreamSummary()
+        tests = []
+
+        def _add_dict(test):
+            tests.append(test)
+
+        outcomes = testtools.StreamToDict(functools.partial(_add_dict))
+        result = testtools.CopyStreamResult([starts, outcomes, summary])
+        result.startTestRun()
+        try:
+            stream.run(result)
+        finally:
+            result.stopTestRun()
+        self.assertThat(len(tests), testtools.matchers.GreaterThan(0))
 
     def assertRunExit(self, cmd, expected, subunit=False, stdin=None):
         if stdin:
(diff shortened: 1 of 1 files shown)


mtreinish:
Heh, that's what I get for only testing locally on py36

mtreinish:
Oh, I actually was hitting this in my os-testr use stestr patch: https://review.openstack.org/#/c/488441/ I was able to workaround that by using the stdout capture fixture there. But, it's actually an upstream bug in subunit. I pushed a patch for it already: https://github.com/testing-cabal/subunit/pull/25 I'll see if I can work around it here too

mtreinish:
Hmm, we're already using the stdout capture fixture here. I'm not sure why sys.stdout is an _io.FileIO object instead of a File object then. That is the crux of the issue that subunit patches is fixing. I'll investigate some more things tomorrow.

mtreinish:
Hah, nice so this exposes a bug in subunit I think. The issue here is having subunit output go to stdout from the test worker is crossing the wires and it's being treated as part of the test run by subunit. This means stestr thinks the fake test fails during test_until_failure_with_subunit_fails_from_func() are real failures.

I thought having it encapsulated in the stdout attachment would mean it's not getting parsed like a normal test run, but apparently not. This is easy enough to workaround, we just redirect stdout to a local buffer. I'll have to investigate if it's a subunit bug.

mtreinish:
The fix for the py36 failures is in #116

stestr/tests/test_return_codes.py
```diff
@@ -199,3 +204,48 @@ def test_load_from_stdin(self):
...
+        self.assertEqual(1, run.run_command(filters=['bad.regex.foobar']))
+
+    def test_parallel_fails_from_func(self):
+        self.assertEqual(1, run.run_command())
+
+    def test_serial_passing_from_func(self):
+        self.assertEqual(0, run.run_command(filters=['passing'], serial=True))
+
+    def test_serial_fails_from_func(self):
+        self.assertEqual(1, run.run_command(serial=True))
```
masayukig: https://travis-ci.org/mtreinish/stestr/jobs/288739848#L673
I'm actually not sure, but this should be like this?
```
stdout = io.BytesIO()
self.assertEqual(1, run.run_command(serial=True, stdout=stdout))
```

masayukig: ah, this causes another issue... :(
```
TypeError: 'unicode' does not have the buffer interface
```

mtreinish: yeah, I've got a branch locally that adds the stdout redirect to everything. I have to debug it some more to work through the unicode errors. But I decided to wait on that until the new subunit release is out, because we're blocked on that anyway.

mtreinish:
Sigh, finally after getting a python-subunit release pushed to fix the py2.7 failures now the windows tests are failing for some reason. This time it looks like a general packaging issue on the windows side: https://ci.appveyor.com/project/mtreinish/stestr/build/1.0.451/job/w47yd53rj980kpel#L152 it seems like tox is failing when calling subprocess to run stestr run. I'm not sure what would have changed to cause this, I guess I'll need to find a windows environment to test this in.

mtreinish:
@masayukig ok, this should finally be good to go now. I finally got it working in all the environments we test and also not leaking to stdout anymore.

masayukig:
Thank you! This patch increases the test coverage. I think there's no reason to hesitate to merge now.

PR #81, opened 2017-08-09
Ensure we always close files in file repository

mtreinish:
In newer versions of python we get a warning if we don't explictly close
an open file. This commit address those warnings by ensuring we always
close any open files by using a context manager.

This was ported from the testrepository patch (which landed soon after
stestr was forked):

https://github.com/testing-cabal/testrepository/pull/25

### DIFF
--- a/.travis/coveralls.sh
+++ b/.travis/coveralls.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 
 if [ "$TOXENV" = "cover" ]; then
-    coverall
+    coveralls
 fi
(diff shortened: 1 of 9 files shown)


mtreinish:
heh, nah I was just doing some git archeology to try and figure out why https://github.com/mtreinish/stestr/commit/7bdfd94f6e3f71269f8f81df4e93d58b214d9ddc doesn't work (I vaguely remember some similar issues from when I first started the project) and I noticed I forked with 1 commit missing, so I ported it over.

If you want feel free to add a commit on top of this branch that fixes the other instances of open without close(). Just be careful because there are some cases where the close call occurs elsewhere in the code. (or outside of stestr's code entirely) We pass around file like objects a lot to manipulate the subunit streams between processes and the repository.

masayukig:
OK, I'll try to do that with care :)

masayukig:
done #84

PR #83, opened 2017-08-09
Add osx test environment to testing

mtreinish:
We already validate that stestr works on windows, but people also use
it from osx. This commit adds the travis config bits necessary to also
run unit tests on an osx environment to verify that we keep things
working there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ matrix:
       env: TOXENV=docs
     - python: "2.7"
       env: TOXENV=py27
+    - os: osx
+      env: TOXENV=py36
+
 cache:
   directories:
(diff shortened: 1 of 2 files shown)


mtreinish:
hah, looks like it's failing on extracting the tarball with the python binary: https://travis-ci.org/mtreinish/stestr/jobs/262738776 I'll have to do some more digging and figure out how difficult this actually is.

mtreinish:
I don't think the osx python support is there, it also seems to be much slower than the linux nodes (probably a more limited pool of resources) At this point it's not worth it for the small bit of extra coverage we get. I might re-open in the future (and if someone else is interested feel free to take this over)

PR #84, opened 2017-08-10
Use context managers for file open()

masayukig:
This commit tries to use context managers for file open(). In the
previous commit, it already has been done as a bug fix. However, these
also can be fixed to close files. In some cases, the original code are
not harmful, but I changed it to keep the code clean.


mtreinish:
fwiw, you probably could have just pushed this commit into the  mtreinish:use-context-managers branch directly and it would have shown up in the PR for that branch. (since you have commit access on the project) But, this works too

masayukig:
heh, yeah, I was thinking about that originally. But this shoul be safe. so, I chose this so far

PR #86, opened 2017-09-08
Update the trove classifiers

mtreinish:
The trove classifiers for stestr are a bit inaccurate, mostly because I
copy and pasted the original setup.cfg from another project. This commit
updates the classifiers in the setup.cfg to better reflect reality.


afrittoli:
Nice  :)

PR #87, opened 2017-09-12
Update MANUAL.rst

cdent:
Fix typo in MANUAL.rst where the black-regex arg was written as back-regex or back-regexp.

black-regex (or -B) is correct


mtreinish:
LGTM, thank!

mtreinish:
Oh, also I just noticed it's missing the run subcommand there too. (I should have looked more closely before the merge) If you want to push a follow up with that I'll review it quickly, otherwise I can do it tomorrow.

PR #89, opened 2017-09-13
Fix: pass the test_path parameter to the 'list' command

ltoscano-rh:


mtreinish:
LGTM, unit tests would have been nice. But this is straightforward enough that it's not a hard requirement.

PR #91, opened 2017-09-14
Make reno an extra requirement

toabctl:
reno is not needed during runtime and also not needed to execute the tests.
It's only needed to build release notes so do not require reno for testing.


mtreinish:
Hmm, so I think actually we can probably actually just completely drop reno. I haven't really been using it or requiring it for changes. There hasn't been really been a release note added in a long time. The other thing is getting reno to actually do something useful outside of openstack infra isn't exactly straightforward. So I think we should drop the releasenote dir and the requirement altogether.

toabctl:
@mtreinish reno dropped.

Issue #88, opened 2017-09-13
output for 'last' command with no .stestr repository is a stacktrace


rm-you:
An error is expected, but the stacktrace it currently outputs is cryptic and unnecessary. Should it just report an error like "No stestr results repository found" or similar?

mtreinish:
Yeah, I think it should just print the error message 'No stestr repository found' and return a non zero exit code.

PR #94, opened 2017-09-15
Print just error message when no stestr repo

masayukig:
This commit makes to print just the error message without a stacktrace
when no stestr repository or no test run. We don't need to show a
stacktrace like this predictable situation.

Closes Issue #88


mtreinish:
LGTM, thanks

PR #60, opened 2017-05-01
Add appveyor ci config

mtreinish:
This commit adds an appveyor config file to run stestr in windows.

PR #96, opened 2017-09-15
Remove pyc files before running stestr in tox

masayukig:
This commit add a removing *.pyc files command before running stestr in
tox. Sometimes, old *.pyc files are remaining. It is not good,
basically.

### DIFF
--- a/tools/find_and_rm.py
+++ b/tools/find_and_rm.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+
+def find_and_remove(suffix='.pyc'):
+    for root, dirs, files in os.walk('.'):
+        for file in files:
+            target = os.path.join(root, file)
+            if os.path.isfile(target) and target.endswith(suffix):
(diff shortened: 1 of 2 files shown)


masayukig:
oh, this shouldn't be approved because of windows test failure.. It doesn't have the unix find command. I need to fix this.

mtreinish:
Yeah this was something I hit when we added the windows ci back in: https://github.com/mtreinish/stestr/pull/60 I didn't have any clever ideas so I just removed the find.

afrittoli:
Sorry! I did this on my mobile and I thought it would just +1 the change. Shall I revert it?

masayukig:
hmm, something wrong.. I just wanted to push the last patch. I'll try it.

masayukig:
I'm trying to remove unnecessary commits..

masayukig:
OK, it's clean now.

tools/find_and_rm.py
```diff
@@ -0,0 +1,31 @@
...
+            if os.path.isfile(target) and target.endswith(suffix):
+                os.remove(target)
+
+
+def main():
+    find_and_remove()
+
+
+if __name__ == '__main__':
+    main()
```
mtreinish: Why not just call find_and_remove() directly here? I don't see what advantage main gives you here

masayukig: ah, yeah, indeed. I'll fix.

mtreinish:
So the code lgtm now, but I'm a bit concerned about the git branch. The way github does merges this will work fine, but my OCD has a problem with landing 3 commits (2 of which are the same) only to have the head of the branch undo that. We really only need 1 commit for this to introduce the script and call it in tox.ini

masayukig:
Hrm, I just push the button "Update branch" on github. But, OK. I'll update and push a patch of 1 commit.

mtreinish:
Yeah, I'm not sure what happened there before. There were a couple merge commits in there which is normally fine to ignore. I was more trying to preserve the ability to bisect changes. There were 2 patches with the red X from earlier attempts. (both adding the same find command) I'm not sure how the branch got into that state though.

PR #98, opened 2017-09-20
Add a --abbreviate flag to stestr run and load

mtreinish:
This commit adds a new flag to the stestr run and load comands for
printing abbreviated output. By default the full subunit-trace output is
used, but in some cases this is overly verbose. To enable a tradeoff
between the full verbosity and the blank output during execution this
flag will print a single character as each test executes.


mtreinish:
My guess is that the unknown args aren't being treated because we call parse_known_args() which makes it non-fatal. (https://docs.python.org/3.4/library/argparse.html#argparse.ArgumentParser.parse_known_args ) See: https://github.com/mtreinish/stestr/blob/master/stestr/cli.py#L94 This is how we get the regexes for selection, but this is really just a bug. The fix is to define a regexes param that has nargs='*' and switch the call to parse_args()

PR #104, opened 2017-09-27
Fix .testr.conf parsing: test path follows discover

ltoscano-rh:
Apparently if -t is not specified, the path to the unit test directory
follows immediately the "discover" keyword.

This change was applied to and tested against the similar code inside
ostestr which parses and coverts .testr.conf on the fly:
https://review.openstack.org/503877/


mtreinish:
LGTM, thanks

Issue #85, opened 2017-08-17
stestr run --analyze-isolation doesn't work


mtreinish:
The --analyze-isolation command used for bisect parallel test failures was never completed from the original fork. There are lots of things broken about that call path that prevent it from even running today. This will all need to be cleaned up. Also because of the complexity of this code path we should add real unit tests to ensure it works properly moving forward.

mtreinish:
I started working on this issue here: https://github.com/mtreinish/stestr/commit/000593f331197c1c7f296c5fb4f3d87318b5b38c

It's still missing real unit tests though. I still need to wrap my head around exactly how the test bisect works so I can write tests for it. The fixes there were fairly mechanical and just code cleanup. But to test it I need to trace through the logic to learn how it works. (which I'm not really in the mental space for right now) If someone else wants to add a patch on top of that branch to add real unit tests and take over the work that would be great. I'm not sure when I'll have time to investigate it further.

mtreinish:
just to update the link in the last comment was for a specific commit, the working branch is: https://github.com/mtreinish/stestr/tree/fix-test-bisect the unit tests still need to be flushed out (and fixed so what's there actually works)

mtreinish:
Now that #107 has landed this can be closed. There are still some UX issues and limitations with the implementation of the flag. But those predate stestr, and are outside of the scope of this bug (which was to fix the flag post stestr pivot) and will need to fixed later. Additional issues should be opened to track those improvements.

PR #107, opened 2017-10-03
Fix the stestr run --analyze-isolation flag

mtreinish:
The analyze-isolation command was never update after the stestr repo was
forked from testrepository. This code path was still making a bunch of
calls assuming the repo was still laid out like testrepository. This
commit fixes all of these and refactors the test biscet code to live in
it's own class. This also adds tests for this logic to ensure it works.

Fixes issue #85


mtreinish:
@afrittoli I think this should fix the bug, but the unit tests might be a bit confusing still

mtreinish:
Hmm, I guess this isn't ready yet then. I'll also have to add unit tests with more than 1 failure so we catch this. I won't be able to look at it until tomorrow though, so if someone wants to tackle that before feel free.

mtreinish:
@masayukig I just pushed an updated revision which I think should address the weird output. I thin it was caused by still using the subunit-trace output for the test_runs, this is problematic because the test bisect does a bunch of test runs. The unit tests don't catch this because I just mock out the runner. I added an additional multi-worker unit test too,

I'm curious how your manually testing this, because I think that would be a good functional test to add. The unit test coverage here only goes so far. If you wanted to add a functional test on top of this branch to add that functional testing I think that would valuable. (you'd have to create a little fake test suite in stestr/tests/files and then run stestr --analyze-isolation on that)

masayukig:
I just tested the latest patch, it seems the weird output is still there... http://paste.openstack.org/show/622804/

> I'm curious how your manually testing this, because I think that would be a good functional test to add. 
heh, I just added assertions like below and run it. And yeah, I agree with that. I'll try.

```diff
diff --git a/stestr/tests/test_return_codes.py b/stestr/tests/test_return_codes.py
index df1164b..a5ac134 100644
--- a/stestr/tests/test_return_codes.py
+++ b/stestr/tests/test_return_codes.py
@@ -156,12 +156,15 @@ class TestReturnCodes(base.TestCase):
     def test_until_failure_with_subunit_fails(self):
         self.assertRunExit('stestr run --until-failure --subunit', 1,
                            subunit=True)
+        self.assertEqual(1, 0)
 
     def test_list(self):
         self.assertRunExit('stestr list', 0)
+        self.assertEqual(1, 0)
 
     def test_no_command(self):
         self.assertRunExit('stestr', 2)
+        self.assertEqual(1, 0)
 
     def _get_cmd_stdout(self, cmd):
         p = subprocess.Popen(cmd, shell=True,
```

afrittoli:
I set up this small example:

```diff
diff --git a/stestr/tests/test_bisect_tests.py b/stestr/tests/test_bisect_tests.py
index 7e9f3be..8a64162 100644
--- a/stestr/tests/test_bisect_tests.py
+++ b/stestr/tests/test_bisect_tests.py
@@ -13,6 +13,7 @@
 import io
 import operator
 import os
+import time
 
 import mock
 import six
@@ -98,3 +99,10 @@ class TestBisectTests(base.TestCase):
             'stestr.tests.test_subunit_trace.TestSubunitTrace.'
             'test_trace_with_all_skips')
         self.assertEqual(['test_that_passed'], prior_tests)
+
+    def test_write_something(self):
+        with open('/Users/andreafrittoli/test_lock', 'a') as lock_file:
+            lock_file.write('Hello, World!')
+        time.sleep(15)
+        os.remove('/tmp/test_lock')
+        pass
diff --git a/stestr/tests/test_config_file.py b/stestr/tests/test_config_file.py
index 4f8b905..1581961 100644
--- a/stestr/tests/test_config_file.py
+++ b/stestr/tests/test_config_file.py
@@ -14,7 +14,7 @@ import mock
 
 from stestr import config_file
 from stestr.tests import base
-
+import testtools
 
 class TestTestrConf(base.TestCase):
 
@@ -56,3 +56,8 @@ class TestTestrConf(base.TestCase):
 
     def test_get_run_command_win32(self):
         self._check_get_run_command()
+
+    def test_assert_not_there(self):
+        with testtools.ExpectedException(FileNotFoundError):
+            with open('/tmp/test_lock', 'r') as lock_file:
+                print(lock_file.readline())
```

With that depending on the concurrency `test_assert_not_there` will fail or succeed, depending on whether the execution happens in parallel to `test_write_something` or not. In isolations all tests will succeed. 

The result is quite not what I was expecting - it's supposed to say something like `failing test', 'caused by test'` and specify the cause or "unknown - no conflicts", but it doesn't say anything at all, only a couple of test runs.

afrittoli:
What I understood so far:
- the  `_prior_tests` helper looks for tests on the same worker that ran prior to the failed test(s). Anything on a different worker will be ignored, so this implementation won't be able to find the culprit in the test case I set-up above. It would only work in case of a test1 that breaks isolation beyond test1 run, so that if test2 runs after test1 it still fails.  I guess it's better than nothing. To find tests across workers I guess we would need to look at timestamps and assume all workers run on the same host or in time-synchronised hosts.
- the run code runs tests to find test that always fail even in isolation. In doing so it creates one or more runs in the repo which hide the run to be analysed from the bisect code, which looks for "last" in the test repo.

afrittoli:
With a few changes I can get the isolation code to run through, and it seems to be doing the right thing now, only it's not what I need for the use case I setup. I'll try to make another one where this kind of bisect is useful.

```diff
diff --git a/stestr/bisect_tests.py b/stestr/bisect_tests.py
index 9b4e9d5..72da60e 100644
--- a/stestr/bisect_tests.py
+++ b/stestr/bisect_tests.py
@@ -31,7 +31,7 @@ class IsolationAnalyzer(object):
         self.repo_type = repo_type
         self.repo_url = repo_url
         self.serial = serial
-        self.concurrency = concurrency,
+        self.concurrency = concurrency
         self.test_path = test_path
         self.top_dir = top_dir
         self.run_func = run_func
@@ -42,9 +42,11 @@ class IsolationAnalyzer(object):
         test_conflicts = {}
         if not spurious_failures:
             raise ValueError('No failures provided to bisect the cause of')
+
         for spurious_failure in spurious_failures:
             candidate_causes = self._prior_tests(self.latest_run,
                                                  spurious_failure)
+
             bottom = 0
             top = len(candidate_causes)
             width = top - bottom
@@ -57,7 +59,7 @@ class IsolationAnalyzer(object):
                     repo_type=self.repo_type, repo_url=self.repo_url,
                     serial=self.serial, concurrency=self.concurrency,
                     test_path=self.test_path, top_dir=self.top_dir)
-                self.run_func(cmd, True, False, False,
+                self.run_func(cmd, False, True, False, False,
                               repo_type=self.repo_type,
                               repo_url=self.repo_url)
                 # check that the test we're probing still failed - still
diff --git a/stestr/commands/run.py b/stestr/commands/run.py
index 57c895f..94cba19 100644
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -310,6 +310,11 @@ def run_command(config='.stestr.conf', repo_type='file',
                               color=color,
                               stdout=stdout)
     else:
+        # Where do we source data about the cause of conflicts.
+        # XXX: Should instead capture the run id in with the failing test
+        # data so that we can deal with failures split across many partial
+        # runs.
+        latest_run = repo.get_latest_run()
         # Stage one: reduce the list of failing tests (possibly further
         # reduced by testfilters) to eliminate fails-on-own tests.
         spurious_failures = set()
@@ -339,11 +344,6 @@ def run_command(config='.stestr.conf', repo_type='file',
         if not spurious_failures:
             # All done.
             return 0
-        # Where do we source data about the cause of conflicts.
-        # XXX: Should instead capture the run id in with the failing test
-        # data so that we can deal with failures split across many partial
-        # runs.
-        latest_run = repo.get_latest_run()
         bisect_runner = bisect_tests.IsolationAnalyzer(
             latest_run, conf, _run_tests, repo, test_path=test_path,
             top_dir=top_dir, group_regex=group_regex, repo_type=repo_type,
```

mtreinish:
@afrittoli @masayukig is there anything else holding this up? I figure we can add functional testing in a follow up patch, but I'd like to get this merged before too long

masayukig:
What do you think @afrittoli ?

afrittoli:
@masayukig yeah I don't like very much how the bisect outputs like if it was a user driven test run, and it appends results into the DB as well (mixing bisect runs with regular runs), but changing that would be a bit more work I think so we should not hold this branch on that.
@mtreinish I would have preferred to have functional tests in but I've been swamped with other things, so I'm fine with merging this and adding tests later on. I wonder if we should include a note or something about this feature being experimental or something?

mtreinish:
Well thinking about the output and repository problems, that will definitely require more work to solve it for real. Mostly because we don't have a method to run tests without storing the results today. (ditto for having some form of output)

In the short term we might be able to leverage the --abbreviate flag or something to tweak subunit-trace to get less output. (I can't remember if there is a method to suppress the stack traces or not) Right now it's not using subunit-trace and falling back to the legacy output stuff which is print nothing except on failure.

What I think we should do is open an issue for the missing features and warts on this, especially because some are more generic features. That way we can track the effort to improve this. Right now I feel like we've got enough to fix #85 because this is more or less what testr did in the past. We can work on improving it from here. (including doing the time stamp analysis to figure out cross worker interactions) As for the functional tests, I'll try to hack something together this afternoon, but I don't want to hold this up really. This is the last thing before we push 1.1.0 out.

mtreinish:
ok I pushed up a new functional tests, I'll look into expanding it in the future. But, this gives us a starting point where we can write fake test suites and then run the commands on them.

mtreinish:
ugh, py3 has different string formatting than py2.

mtreinish:
haha, well I guess the windows testing is paying off.

masayukig:
Oh, what's happening on the windows?

mtreinish:
I'm not entirely sure yet, but the functional tests are returning 0 and not 3 when we run --analyze-isolation: https://ci.appveyor.com/project/mtreinish/stestr/build/1.0.258/job/5g5gy7rgmuaifwi5#L770

mtreinish:
after adding some more debug output, it looks like subunit is doing a select() call:

https://ci.appveyor.com/project/mtreinish/stestr/build/1.0.259/job/mspcp3t50yrd8rja#L851

That will be problematic on windows, since select() isn't a thing there...

mtreinish:
Ah this is what @bclau fixed on subunit with: https://github.com/testing-cabal/subunit/commit/7149380ff9a4f77f523b4b82e88b22c4e55a77a6

mtreinish:
@masayukig and @afrittoli I'm thinking I'm just going to remove the functional tests from this PR for now. It's blocked on a new subunit release which may take some time. I'd rather land this and get the next stestr release out the door and land the extra testing a bit later. Any issue with doing that?

afrittoli:
@mtreinish Is it an option to skip the functional test on windows? I'd rather have a little functional coverage, since you did the work already, than none.

mtreinish:
That's a good point, I'll add a skip condition to the test

mtreinish:
@masayukig and @afrittoli ok, skipping the single functional test on windows everything is passing now. I say lets merge this now and we can expand the testing in the future and also fix the UI warts.

afrittoli:
@mtreinish @masayukig lgtm!

Issue #99, opened 2017-09-21
No error handling for unknown arguments


mtreinish:
Unknown args aren't being treated as fatal because we call parse_known_args() which makes it non-fatal. (https://docs.python.org/3.6/library/argparse.html#argparse.ArgumentParser.parse_known_args ) See: https://github.com/mtreinish/stestr/blob/master/stestr/cli.py#L94 

We use this method to get the regexes for selection, but this is really just a bug. The fix here is to just define a regexes param that has nargs='*' and switch the call to parse_args().

This was originally found by @masayukig in #98

masayukig:
I made a pull request for using cliff #100 which resolving this issue. But if you want to resolve only this asap, I can take this.

masayukig:
This can be closed with #100 merging.

mtreinish:
Yep, let's close this:

$ stestr run --asd
usage: stestr run [-h] [--failing] [--serial] [--concurrency CONCURRENCY]
                  [--load-list LOAD_LIST] [--partial] [--subunit]
                  [--until-failure] [--analyze-isolation] [--isolated]
                  [--worker-file WORKER_PATH]
                  [--blacklist-file BLACKLIST_FILE]
                  [--whitelist-file WHITELIST_FILE]
                  [--black-regex BLACK_REGEX] [--no-discover TEST_ID]
                  [--random] [--combine] [--no-subunit-trace]
                  [--force-subunit-trace] [--color] [--slowest] [--abbreviate]
                  [filters [filters ...]]
stestr run: error: unrecognized arguments: --asd

PR #100, opened 2017-09-22
Use cliff for CLI layer

masayukig:
This commit makes stestr use cliff[1] for CLI layer. The cliff project
provides a lot of advantages for stestr cli which has subcommands.
Instead of just using argparse, we should leverage cliff to provide a
more plished CLI experience.

[1] https://pypi.python.org/pypi/cliff

Closes Issue #62

PR #113, opened 2017-10-16
Fix error handling for unknown arguments

masayukig:
This commit fixes to handle unknown arguments. Unknown arguments should
be treated as fatal. Otherwise, users can't know their options are valid
or not at the runtime.

Fixes issue #99

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -48,8 +48,7 @@ three functions::
          real work. The return value is expected to be an int and is used for
          the exit code (assuming the function doesn't call sys.exit() on it's
          own)"""
-      args = arguments[0]
-      unknown_args = arguments[1]
+      args = arguments
       return call_foo()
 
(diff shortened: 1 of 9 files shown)


mtreinish:
@masayukig This looks fine to me, but did you want to land this to fix #99 for the next release? I thought you were planning to just fix this with the cliff migration in #100 (which btw has the best commit number)

masayukig:
Ah, I'm fine with fixing this with the cliff migration :) yeah, I agree with the commit number is the best.

mtreinish:
ok, so what's the path forward here, do you want to land this now, push 1.1.0, and then get #100 landed? Or just abandon this, wait for 1.1.0, and then only do #100?

masayukig:
well, I was thinking just abandon this, and wait for 1.1.0. I actually don't know how to abandon, though..

masayukig:
ah, I found that this "Close and comment" button is for it :)

Issue #115, opened 2017-10-16
stestr load (and run by extension) will unconditionally write stdout to sys.stdout


mtreinish:
The stestr load command function has a kwarg stdout to specify the file object to write to for stdout. However in the case when subunit_out is set to true this is not honored and it unconditionally writes to sys.stdout instead of the provided file.

PR #116, opened 2017-10-16
Pass output to output.make_result() in stestr load

mtreinish:
The stestr load command was not passing the stdout arg to the
make_result call, which is what we use to generate the subunit output.
This commit fixes this oversight and passing stdout to the function to
make sure we write the subunit output to the specified file instead of
unconditionally using sys.stdout.

Fixes Issue #115


mtreinish:
The tests for this are in #79 (that's how I found the bug). That is being blocked on py27 by a subunit bug, but will merge as soon as there is a new subunit release.

Issue #108, opened 2017-10-05
Deprecate the --partial options


mtreinish:
Tracing through the code the use of --partial on _stestr run_ and _stestr load_ don't seem to have any real effect. This option was directly ported from testr in the initial fork and I didn't really think about it but the flags serve no real purpose from what I can tell. I think the original idea was to distinguish an entry in a repository for a full run vs one that was a subset. (but you'd have to check with the testrepository authors on that) But nothing else respects or expects that difference, and from an end user perspective it doesn't really matter. The repository contains the historical record of previous runs and information for scheduling is aggregated independently from that. So it doesn't really matter if the record is for a partial execution or not.

Therefore I think it'll be good to just remove those options, Unfortunately the options are in our stable interfaces (both the cli and python commands api) so we'll have to deprecate them for a bit first. We can remove the flags and just print a deprecation warning on usage to warn people things are going away.

The only functional code that seems to use it is in the repository:

https://github.com/mtreinish/stestr/blob/86b8ee4709538e18b72a7373ce9dad19c5015a12/stestr/repository/file.py#L314-L322

and 

https://github.com/mtreinish/stestr/blob/82b3efc3f4d872d04424b8cdfd466942dbf95d92/stestr/repository/memory.py#L163-L164

Which are related, the file repo uses the memory repo call after that highlighted if statement. I'm not exactly sure on the function of that code so investigation will be needed to understand what's being triggered there and how to implement it without a user facing flag.

masayukig:
oh, github doesn't support the "Partial-Bug" keyword :-p

mtreinish:
heh, I don't think github supports any keywords except for the '#' references. I haven't found any automation around closing issues. Although I haven't bothered to read any of the documentation on it.

mtreinish:
@masayukig do you think we can close this since we landed the deprecation patch? Or do you think we should wait until the removal patch at some point in the future (it'll probably be a long time because it's an annoying breaking change)

masayukig:
Yeah, I prefer closing this issue and make a new issue for removing the --partial option. And it's better to change the title of this issue to like "Deprecate the --partial options".

PR #117, opened 2017-10-17
Deprecate the partial flag

mtreinish:
The partial flag is vestigial and left over from the original testr fork.
It was originally used to indicate a run did not execute the entire test
suite. However this doesn't really come into play anywhere else and
it's not stored in the repository either. I think the eventual intent
was to treat those records as incomplete for other purposes but that
never really was implemented. With the introduction of load to run-id
and also a run --combine this isn't really needed anymore either. It's not
really something that should be user facing either. The user shouldn't
have to tell the runner whether the run is the full suite or not.

This commit removes the functional bits from the cli (and their python
api commands) and deprecates all the flags for partial. The repository
partial usage remains intact because it's coupled to how the file
repository runs. (it uses a memory repository with partial=True before
writing it to disk) So more care is needed before we rip that out, but
in the mean time this removes and deprecates all the end user facing
pieces.

Closes issue #108

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -15,6 +15,7 @@
 import os
 import subprocess
 import sys
+import warnings
 
 import six
 import subunit
@@ -42,8 +43,10 @@ def set_cli_opts(parser):
     parser.add_argument("--load-list", default=None,
                         help="Only run tests listed in the named file."),
     parser.add_argument("--partial", action="store_true", default=False,
-                        help="Only some tests will be run. Implied by "
-                             "--failing.")
+                        help="DEPRECATED: Only some tests will be run. Implied"
+                             " by --failing. This option is deprecated and no "
+                             "longer does anything. It will be removed in the "
+                             "future")
     parser.add_argument("--subunit", action="store_true", default=False,
                         help="Display results in subunit format.")
     parser.add_argument("--until-failure", action="store_true", default=False,
@@ -159,7 +162,9 @@ def run_command(config='.stestr.conf', repo_type='file',
         autodetects your CPU count and uses that.
     :param str load_list: The path to a list of test_ids. If specified only
         tests listed in the named file will be run.
-    :param bool partial: Only some tests will be run. Implied by `--failing`.
+    :param bool partial: DEPRECATED: Only some tests will be run. Implied by
+        `--failing`. This flag is deprecated because and doesn't do anything
+        it will be removed in a future release.
     :param bool subunit_out: Display results in subunit format.
     :param bool until_failure: Repeat the run again and again until failure
         occurs.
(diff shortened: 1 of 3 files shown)


mtreinish:
ooh, nice a .006% coverage increase :p

stestr/commands/run.py
```diff
@@ -42,8 +43,10 @@ def set_cli_opts(parser):
     parser.add_argument("--load-list", default=None,
                         help="Only run tests listed in the named file."),
     parser.add_argument("--partial", action="store_true", default=False,
-                        help="Only some tests will be run. Implied by "
-                             "--failing.")
+                        help="DEPRECATED: Only some tests will be run. Implied"
+                             " by --failing. This option is deprecated and no "
+                             "longer does anything. It will be removed in the"
+                             "future")
```
masayukig: Do you want to add a white space before `future`?

mtreinish: oh, good catch. I'll respin this in a bit

mtreinish:
Yeah, I saw the same thing in the docs and was debating on that, I'm not sure which is better. This is a warning for deprecation, and I don't necessarily want to spam people with it. But I do want it to be discoverable which this isn't. What do you think?

mtreinish:
Heh, well I was asking your opinion there. I don't actually know what is best here, I just read the warning module docs and that seemed to be what we wanted, because this is a deprecation. But, it might not be best. Maybe it's better to just not set a warning type and use the default so it'll be printed. I don't feel strongly either way.

Yeah, --partial shouldn't ever really be used it didn't really do anything so either way it's fine. It's more about signaling our users that the flag is going away so they'll need to update their usage.

masayukig:
I actually expected the warnings would be printed. But, I think we need to write a bit more code to print them. I don't think it's worth doing that.

mtreinish:
Well I think it's the warning class there. On py3 I know they print regular warnings by default. I can switch it to a print too. Let me test some things locally and figure out the best compromise here

mtreinish:
ok, I just removed the deprecation warning class type from the warn() call. In my local testing it prints once to stdout when you use the --partial flag. Granted it's not the most clean, but that's because upstream python never thinks about the end user experience. Do you think I should just switch it to a print?

stestr/commands/run.py
```diff
@@ -194,6 +199,9 @@ def run_command(config='.stestr.conf', repo_type='file',
         for failures.
     :rtype: int
     """
+    if partial:
+        warnings.warn('The partial flag is deprecated and has no effect  '
```
masayukig: heh, there are two white spaces after the word 'effect'.

mtreinish:
Ok, the latest rev quickly fixes the double space, I'll merge it after the CI jobs come back.

Issue #121, opened 2017-10-20
Add a silent output mode to stestr run and stestr load


mtreinish:
There is an occasional use case for having a flag on stestr run and load that doesn't output anything. It just returns either 0 or 1. We should add a flag to stestr run and load that does this.

The primary motivator for this is the --analyze-isolation flag which does multiple test runs during it's operation. If there are failures they print to stdout making it confusing to follow.

masayukig:
Is the '-q'/'--quiet' option for this mode? It seems that it doesn't work now, though.

```
  -q, --quiet           Turn off output other than the primary output for a
                        command and any errors.
```

mtreinish:
Yeah, I don't think I ever did anything with that besides the definition. IIRC it''s defined in the base cli module, we probably can reuse the definition if we want, but it's more the implementation details that matter for this bug.

masayukig:
How about like this? #124

mtreinish:
#124 landed so we can close this

PR #124, opened 2017-10-23
Implement quiet option for run and load commands

masayukig:
This commit makes run and load commands use the quiet option. We don't
need outputs of run and load commands occasionally. So, this patch adds
the function to suppress the outputs.

We already have the quiet option, however it isn't used so far. It might
be better to move the quiet (or silent) option to load and run commands
because the option doesn't work in the other commands now.

Fixes issue #121

### DIFF
--- a/stestr/commands/load.py
+++ b/stestr/commands/load.py
@@ -15,6 +15,7 @@
 
 import datetime
 import functools
+import os
 import sys
 import warnings
 
@@ -68,11 +69,12 @@ def get_cli_help():
 
 def run(arguments):
     args = arguments[0]
+    stdout = open(os.devnull, 'w') if args.quiet else sys.stdout
     load(repo_type=args.repo_type, repo_url=args.repo_url,
          partial=args.partial, subunit_out=args.subunit,
          force_init=args.force_init, streams=arguments[1],
          pretty_out=args.subunit_trace, color=args.color,
-         abbreviate=args.abbreviate)
+         stdout=stdout, abbreviate=args.abbreviate)
 
 
 def load(force_init=False, in_streams=None,
--- a/stestr/results.py
+++ b/stestr/results.py
@@ -154,7 +154,7 @@ class CLITestResult(testtools.StreamResult):
             values.append(('skips', skips, None))
         output.output_summary(
             not bool(failures), self._summary.testsRun, num_tests_run_delta,
-            time, time_delta, values)
+            time, time_delta, values, output=self.stream)
 
     def startTestRun(self):
         super(CLITestResult, self).startTestRun()
--- a/stestr/tests/test_return_codes.py
+++ b/stestr/tests/test_return_codes.py
@@ -73,6 +73,7 @@ class TestReturnCodes(base.TestCase):
             self.assertEqual(
                 p.returncode, expected,
                 "Stdout: %s; Stderr: %s" % (out, err))
+            return (out, err)
         else:
             self.assertEqual(p.returncode, expected,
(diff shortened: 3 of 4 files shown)


stestr/commands/load.py
```diff
@@ -68,11 +69,12 @@ def get_cli_help():
 
 def run(arguments):
     args = arguments[0]
+    stdout = open(os.devnull, 'w') if args.quiet else sys.stdout
```
mtreinish: Heh, this is a bit hackier than what I was imagining. I was assuming just to not run the output bits at all if the flag was set, but I guess this is fine way to do it too. My only question is does this work on windows?

mtreinish:
We probably should also add some unit tests or something to verify there is no output but the return codes work.

stestr/commands/load.py
mtreinish: I can answer the windows question for myself by reading the docs: https://docs.python.org/2/library/os.html#os.devnull

masayukig: heh, yeah, python3 as well. I think it should work.
https://docs.python.org/3/library/os.html#os.devnull

masayukig:
Yes, I agree with the adding some unit tests. We probably can verify `os.devnull` can work in Windows environments with them.

stestr/tests/test_return_codes.py
```diff
@@ -199,3 +201,16 @@ def test_load_from_stdin(self):
...
+        out, err = self.assertRunExit('stestr -q run passing', 0)
+        self.assertEqual(out.decode('utf-8'), '')
+        # FIXME(masayukig): We get some warnings when we run a coverage job.
+        # So, just ignore 'err' here.
+        stream = self._get_cmd_stdout(
+            'stestr last --subunit')[0]
+        out, err = self.assertRunExit('stestr -q load', 0, stdin=stream)
+        # FIXME(masayukig): I expected the 'out' was empty, however, we get
+        # some standard outputs, actually such as
+        # 'Ran 79 tests in 5.930s\nPASSED (id=150)'.
```
mtreinish: I feel like this is a bug, we should be passing the stdout variable to all the output helper functions. If something is leaking to stdout then we're probably missing that somewhere.

masayukig: heh, yeah, I'll dig it. A unit test is important :)

stestr/results.py
```diff
@@ -154,7 +154,7 @@ def _output_summary(self, run_id):
             values.append(('skips', skips, None))
         output.output_summary(
             not bool(failures), self._summary.testsRun, num_tests_run_delta,
-            time, time_delta, values)
+            time, time_delta, values, output=self.stream)
```
masayukig: I found that this is the missing part of output. The default of output is sys.stdout.

stestr/tests/test_return_codes.py
```diff
@@ -73,6 +73,7 @@ def assertRunExit(self, cmd, expected, subunit=False, stdin=None):
             self.assertEqual(
                 p.returncode, expected,
                 "Stdout: %s; Stderr: %s" % (out, err))
+            return (out, err)
```
mtreinish: Part of me feels like this is not a good way to do this. I always think an assert function shouldn't have a return at all. It either works and does nothing or fails and explodes. But I guess this is fine for now, it's probably better than calling subprocess by hand 2 or 3 times in the test function and doing the asserts manually. Maybe as a follow up change we can extract the subprocess piece into a separate private function and have the assertMethod call that for subprocessing and we can remove the return here. Because we can use that new method in the test directly.

PR #127, opened 2017-11-19
Use binary type for all subunit data

mtreinish:
In the ReturnCodeToSubunit wrapper class we process the subunit stream
input and handle non zero return codes to ensure worker failures are
tracked. However, subnit data is always binary and not a string, so we
should ensure all data we add to the stream is binary type. There was
one incorrect cast in the read() method using six.text_type() which is
the incorrect type to use. This commit fixes this by switching it to be
six.binary_type() instead.

### DIFF
--- a/stestr/output.py
+++ b/stestr/output.py
@@ -197,7 +197,7 @@ class ReturnCodeToSubunit(object):
 
     def read(self, count=-1):
         if count == 0:
-            return six.text_type('')
+            return six.binary_type('')
         result = self.source.read(count)
         if result:
             self.lastoutput = result[-1]


mtreinish:
TBH, looking at the code for six I don't think this makes any difference except in python2 where it will be an empty unicode string (instead of a normal string which is bytes). On python3 there really isn't any difference between str('') and bytes(''), not that there is any real difference on python2 except for the u'' being printed. This was more just a feel good thing, and also like a 1% chance it would fix things.

mtreinish:
Also, for comparison the equivalent code here from testr: https://github.com/testing-cabal/testrepository/blob/master/testrepository/commands/run.py#L92-L100 is more or less the same except that _b() is testtools.compat._b() which on python3 (python 2 has no change) just returns the 'latin-1' encoding: https://github.com/testing-cabal/testtools/blob/master/testtools/compat.py#L60-L62

stestr/output.py
```diff
@@ -197,7 +197,7 @@ def _append_return_code_as_test(self):
 
     def read(self, count=-1):
         if count == 0:
-            return six.text_type('')
+            return six.binary_type('')
```
masayukig: I guess this occurs an error in py3 like below..

```
>>> six.text_binary('')
  File "<stdin>", line 1, in <module>
AttributeError: module 'six' has no attribute 'text_binary'
```

masayukig: oh, I misunderstood... :(

mtreinish:
Meh, this isn't really needed anyway, so I'll just close this.  I think that error just shows we never actually hit that code path anywhere (either in the real world, or in the unit tests). Which makes sense because when would something call read(0) on the subunit stream?

masayukig:
yeah, I'm fine with closing this. I was trying to write a unit test for this like this, actually.

```
    def test_return_code_to_subunit_read(self):
        p = subprocess.Popen("echo ''", shell=True,
                             stdout=subprocess.PIPE)
        s = output.ReturnCodeToSubunit(p)
        r = s.read(0)
        self.assertEqual(r, '')
```

PR #129, opened 2017-11-27
list: Deal with [] instead of 'None' for filters

javacruft:
argparse returns an empty array for any unknown arguments; detect
this situation and default filters to None in this instance as
already performed in the 'run' cli module.

### DIFF
--- a/stestr/commands/list.py
+++ b/stestr/commands/list.py
@@ -48,7 +48,7 @@ def set_cli_opts(parser):
 
 def run(arguments):
     args = arguments[0]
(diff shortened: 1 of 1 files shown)


mtreinish:
Seems straightforward enough, looking at https://github.com/mtreinish/stestr/blob/master/stestr/commands/list.py#L104 we check if it's None. So when running from the CLI that check will never eval to be true. It does look like we do the same call inside the TestProcessorFixture we create there, but that path isn't exactly straightforward.

PR #131, opened 2017-12-11
Add --slowest option to test runs

stmcginnis:
ostestr provided a --slowest option that made it easy to get
a list of the slowest tests run. This is still possible with
`stestr slowest`, but there are cases where it would be more
convenient to have this automatically printed out without the
need of a separate command call.

This patch adds a --slowest option to the run, defaulted to
False, that will perform the slowest test output as part of
the run command.


mtreinish:
LGTM, thanks!

PR #132, opened 2017-12-21
Fix typos in docs

masayukig:
This commit fixes some typos in the docs. And this commit removes
'search' link on the index page because the page isn't rendered on the
readthedocs page[0] and we can search the docs with the search box on
the left side on every page.

[0] http://stestr.readthedocs.io/en/latest/search.html


mtreinish:
LGTM, thanks

PR #133, opened 2017-12-21
Remove unnecessary shebang and execution bit from subunit_trace

masayukig:
This subunit_trace.py was copied from os_testr[0] which has
"subunit_trace" command in it. However, stestr doesn't have that
command. So the shebang and execution bit aren't necessary. So, this
commit just removes them to keep the code simple and clean.

### DIFF
--- a/stestr/subunit_trace.py
+++ b/stestr/subunit_trace.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
-
 # Copyright 2014 Hewlett-Packard Development Company, L.P.
 # Copyright 2014 Samsung Electronics
 # All Rights Reserved.


mtreinish:
Hmm, I actually left that on purpose when I added subunit-trace as the default output filter. I was debating adding the console script entrypoint to make it a stand alone callable like what os-testr packages. The problem with that is it would conflict on set ups with os-testr.  Do you think we should just abandon that goal of packaging it as a command and leave it as strictly a module in stestr?

masayukig:
oh, I see the point. IIUC, os-testr's subunit-trace will be kept in that package in the future. So, we shouldn't have subunit-trace command in stestr package because of the conflict as you said.
So, yeah, I think we should just abandon that goal and leave it as a module in stestr.

mtreinish:
ok, that works for me. If we want to make it a standalone command in the future we can always easily add this back too.

PR #139, opened 2018-02-14
Add support for a user config file

mtreinish:
This commit adds a new feature, the user config file. This enables users
to write a global config file for UI preferences on certain commands
that will be the new default for that option. For example, if you prefer
the default output on the 'stestr run' command to be with the
--no-subunit-trace flag set. You can just write a user config file to
set this output. Then for all stestr run invocations the default output
will be with --no-subunit-trace set.

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -354,6 +354,42 @@ There is also an option on ``stestr run``, ``--random``/``-r`` to randomize the
 order of tests as they are passed to the workers. This is useful in certain
 use cases, especially when you want to test isolation between test cases.
 
+
+User Config Files
+-----------------
+
+If you prefer to have a different default output or setting for a particular
+command stestr enables you to write a user config file to overide the defaults
+for some options on some commands. By default stestr will look for this config
+file in ``~/.stestr.yaml`` and ``~/.config/stestr.yaml`` in that order. You
+can also specify the path to a config file with the ``--user-config``
+parameter.
+
+The config file is a yaml file that has a top level key for the command and
+then a sub key for each option. For an example, a fully populated config file
+that changes the default on all available options in the config file is::
+
+    run:
+      concurrency: 42 # This can be any integer value >= 0
+      random: True
+      no-subunit-trace: True
+      color: True
+      abbreviate: True
+      slowest: True
+    failing:
+      list: True
+    last:
+      no-subunit-trace: True
+      color: True
+    load:
+      force-init: True
+      subunit-trace: True
+      color: True
+      abbreviate: True
+
+If you choose to use a user config file you can specify any subset of the
+options and commands you choose.
+
 Automated test isolation bisection
 ----------------------------------
 
--- a/stestr/cli.py
+++ b/stestr/cli.py
@@ -48,6 +48,11 @@ class StestrCLI(app.App):
         return parser
 
     def _set_common_opts(self, parser):
+        parser.add_argument('--user-config', dest='user_config', default=None,
+                            help='An optional path to a default user config '
+                                 'file if one is not specified ~/.stestr.yaml '
+                                 'and ~/.config/stestr.yaml will be tried in '
+                                 'that order')
         parser.add_argument('-d', '--here', dest='here',
                             help="Set the directory or url that a command "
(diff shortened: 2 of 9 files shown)


mtreinish:
I will be amending this to add unit test coverage in a bit, we probably shouldn't merge this until then. I created the pull request before that to get initial feedback.

mtreinish:
Ok, I added unit test coverage for the new user config parsing module. I'd like to add coverage for the CLI with a user config file, but we don't have a framework in repo that makes that convenient. I expect the framework for that is going to be pretty big, and this PR is big enough. So I'll save that as a future todo, definitely something we'll need to land before we include this feature in a release.

mtreinish:
@emonty @zaneb I think this is a feature you'll like

zaneb:
Nice, thanks Matt!

emonty:
This is everything I've ever wanted in life.

stestr/cli.py
```diff
@@ -48,6 +48,11 @@ def build_option_parser(self, description, version, argparse_kwargs=None):
         return parser
 
     def _set_common_opts(self, parser):
+        parser.add_argument('--user-config', dest='user_config', default=None,
+                            help='An optional path to a default user config '
+                                 'file if one is not specified ~/.stestr.yml '
+                                 'and ~/.config/stestr.yml will be tried in '
```
mtreinish: Oops, this should be ~/.stestr.conf and ~/.config/.stestr.conf (instead of .yml) that's what is implemented in code. I'll respin this quickly right now.

doc/source/MANUAL.rst
```diff
@@ -354,6 +354,42 @@ There is also an option on ``stestr run``, ``--random``/``-r`` to randomize the
...
 use cases, especially when you want to test isolation between test cases.
 
+
+User Config Files
+-----------------
+
+If you prefer to have a different default output or setting for a particular
+command stestr enables you to write a user config file to overide the defaults
+for some options on some commands. By default stestr will look for this config
+file in ``~/.stestr.conf`` and ``~/.config/stestr.conf`` in that order. You
```
masayukig: I feel this is a bit confusing naming. I was thinking the format would be "INI" which is same as ``.stestr.conf``. But this is a yaml file. Why not use .stestr.yaml/stestr.yaml here?

stestr/commands/run.py
```diff
@@ -127,30 +132,55 @@ def get_description(self):
...
+            if not user_conf.run.get('no-subunit-trace'):
+                if not args.no_subunit_trace:
+                    pretty_out = False
+                else:
+                    pretty_out = True
+            else:
+                pretty_out = False
+
+            pretty_out = args.force_subunit_trace or pretty_out
+            concurrency = args.concurrency or user_conf.run.get(
```
masayukig: I think these don't work correctly because if a user specify 0 or False for concurrency, random, color and etc. with the command line, this code just ignores them.

Probably, we need to remove the default values from add_argument() to handle here correctly?

mtreinish: So I think the logic works out fine for the binary store_true flags because there isn't a way to explicitly say use the flag but as a value that evals false. The flag is either true and used or the default and false.

But you're right, with concurrency a user could manually specify 0 and we'd want that to take precedence over the user config file default. I'll have to rework the logic (and the default) a bit to handle that edge case.

doc/source/MANUAL.rst
mtreinish: I can change it back to yaml, I don't have a strong preference one way or the other.

mtreinish:
@masayukig ok, I've revised the filenames and also updated the logic for the concurrency flag to handle the case where 0 is passed in on the cli. At the same time I caught a bug in the pretty_out logic where I had things backwards which is now fixed too. (when a config file existed but no-subunit-trace was not specified in it)

mtreinish:
Sure, there's no rush. Also if you could take notes on the cases you're manually testing, it'll be good to add  those as functional tests after this merges.

masayukig:
I've tested some manually. What I did is "sub-command + options" random combination testing like `stestr run --slowest; stestr last --color; ...etc.`.

I found no issues except #140 . So, I think we can merge this.

Issue #140, opened 2018-02-20
"'Namespace' object has no attribute 'repo_type'" when running with `--slowest` option


masayukig:
When I run the command like below, I got an error `'Namespace' object has no attribute 'repo_type'` :

```
$ stestr --debug run --slowest
<SNIP>
'Namespace' object has no attribute 'repo_type'
Traceback (most recent call last):
  File "/home/masayuki/git/stestr/.tox/py36/lib/python3.6/site-packages/cliff/app.py", line 400, in run_subcommand
    result = cmd.run(parsed_args)
  File "/home/masayuki/git/stestr/.tox/py36/lib/python3.6/site-packages/cliff/command.py", line 184, in run
    return_code = self.take_action(parsed_args) or 0
  File "/home/masayuki/git/stestr/stestr/commands/run.py", line 154, in take_action
    slowest.slowest(repo_type=args.repo_type, repo_url=args.repo_url)
AttributeError: 'Namespace' object has no attribute 'repo_type'
Traceback (most recent call last):
  File "/home/masayuki/git/stestr/.tox/py36/bin/stestr", line 10, in <module>
    sys.exit(main())
  File "/home/masayuki/git/stestr/stestr/cli.py", line 96, in main
    return cli.run(argv)
  File "/home/masayuki/git/stestr/.tox/py36/lib/python3.6/site-packages/cliff/app.py", line 279, in run
    result = self.run_subcommand(remainder)
  File "/home/masayuki/git/stestr/.tox/py36/lib/python3.6/site-packages/cliff/app.py", line 400, in run_subcommand
    result = cmd.run(parsed_args)
  File "/home/masayuki/git/stestr/.tox/py36/lib/python3.6/site-packages/cliff/command.py", line 184, in run
    return_code = self.take_action(parsed_args) or 0
  File "/home/masayuki/git/stestr/stestr/commands/run.py", line 154, in take_action
    slowest.slowest(repo_type=args.repo_type, repo_url=args.repo_url)
AttributeError: 'Namespace' object has no attribute 'repo_type'
```

https://github.com/mtreinish/stestr/blob/master/stestr/commands/run.py#L154
`self.app_args....` should be passed instead of `args....` here.

mtreinish:
#141 has been merged fixing this

PR #141, opened 2018-02-21
Fix no attribute 'repo_type' error when running w/ --slowest option

masayukig:
This commit fixes the issue "no attribute `repo_type` error when
running with `--slowest` option. We have neither `repo_type` nor
`repo_url` in `args` but `self.app_args`.

Fixes Issue #140


mtreinish:
LGTM, thanks

PR #142, opened 2018-02-21
Add functional tests for user config file

mtreinish:
This commit adds additional testing to cover the newly added user config
file feature from #139. This includes various permutations of CLI and
config file options to test for the correct operational behavior. This
testing managed to uncover a bug with empty config files which was
fixed.

One outstanding issue is the load abbreviated output option which
doesn't fully function. The abbreviated output (ie the '....') doesn't
get displayed, other than that it's fully functional. I assume this is a
bug in the subunit_trace module, so for now those 2 tests are skipped.

### DIFF
--- a/stestr/tests/test_user_config_return_codes.py
+++ b/stestr/tests/test_user_config_return_codes.py
@@ -1,0 +1,444 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+import functools
+import io
+import os
+import shutil
+import subprocess
+import tempfile
+
+import six
+from six import StringIO
+import subunit as subunit_lib
+import testtools
+import yaml
+
+from stestr.tests import base
+
+
+class TestReturnCodes(base.TestCase):
+    def setUp(self):
+        super(TestReturnCodes, self).setUp()
+        # Setup test dirs
+        self.directory = tempfile.mkdtemp(prefix='stestr-unit')
+        self.addCleanup(shutil.rmtree, self.directory)
+        self.test_dir = os.path.join(self.directory, 'tests')
+        os.mkdir(self.test_dir)
+        # Setup Test files
+        self.testr_conf_file = os.path.join(self.directory, '.stestr.conf')
+        self.setup_cfg_file = os.path.join(self.directory, 'setup.cfg')
+        self.passing_file = os.path.join(self.test_dir, 'test_passing.py')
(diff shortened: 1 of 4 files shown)


mtreinish:
sigh, windows...

mtreinish:
ok, I think this latest rev should fix it on windows with python 2.7. My guess is for some reason setting buffering to 1 instead of 0 was not actually flushing the buffer on windows with 2.7 (which it was on py3.5 and py3.6) despite writing more than 1 byte to the file. So the tests were running with an empty user config file and failing. I fixed it by switching all the config file writes to be unbuffered binary and that seems to be working on windows with py2.7 now.

masayukig:
yeah, it seems the reporting/coverage collection doesn't work with subprocessed testing correctly..

masayukig:
And, I'm wondering if you could see source on the coverage page

https://coveralls.io/builds/15634193/source?filename=stestr%2Fuser_config.py#L69
```
Source Not Available

The owner of this repo needs to re-authorize with github; their OAuth credentials are no longer valid so the file cannot be pulled from the github API.
```

stestr/tests/test_user_config_return_codes.py
```diff
@@ -0,0 +1,444 @@
...
+                'last': {
+                    'no-subunit-trace': True,
+                },
+                'load': {
+                    'abbreviate': True,
+                }
+            }, default_flow_style=False))
+        conf_file.write(contents.encode('utf-8'))
+        # NOTE(mtreinish): Running serially here to ensure a consistent
+        # execution order for confirming the abbreviated output.
```
masayukig: I'm curious here. Can we get always the same order with the serial running? I didn't look at the doc/code yet, though.

stestr/tests/test_user_config_return_codes.py
```diff
@@ -0,0 +1,444 @@
...
+        out, err = self.assertRunExit('stestr --user-config=%s load' % path,
+                                      0, stdin=stream)
+        out = six.text_type(out)
+        self.assertNotIn('FAILED (id=0, failures=2)', out)
+        self.assertNotIn('FF..', out)
+        self.assertIn('Totals', out)
+        self.assertIn('Worker Balance', out)
+        self.assertIn('Sum of execute time for each test:', out)
+
+    @testtools.skip('Abbreviated output not displaying')
```
masayukig: OK, we'll revisit here.

stestr/tests/test_user_config_return_codes.py
masayukig: As far as in my local test, the order is the same every time.

masayukig:
As far as my local testing manually, I couldn't find any issues except you already mentioned. So, I think we can release 2.0.0 if you want.

mtreinish:
@masayukig https://coveralls.io/builds/15634193/source?filename=stestr%2Fuser_config.py#L69 loads for me on my desktop. But, when I looked at it on my phone it didn't display and had the same error you're encountering.

mtreinish:
ok, I'll push out the release today then

mtreinish:
As for the ordering issue, it will be deterministic in how it distributes the 4 tests to the workers. But there is no guarantee on how quickly each worker will execute the test since they're starting at more or less the same time. In my testing it was normally consistent, but occasionally the order would change (probably based on system load or core count). I made it run those tests in the subprocess serially just to remove the chance of that happening.

PR #143, opened 2018-02-23
Update the package summary in setup.cfg metadata

mtreinish:
This commit updates the description of the stestr package. The one in
setup.cfg is quite old back from when we were still deciding on the
scope of the project. We have since updated the descriptions everywhere
else to explain the scope of the project. The setup.cfg summary is what
gets sent to pypi: https://pypi.python.org/pypi/stestr/2.0.0 so we
should update this to make sure it reflects the current scope of the
project more accurately.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = stestr
-summary = A test runner runner similar to testrepository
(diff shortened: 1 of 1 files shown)


mtreinish:
@masayukig this is too late for the 2.0.0 release, but at least it will be correct moving forward

PR #151, opened 2018-03-07
Add usage to manual docs page

mtreinish:
This commit leverages the cliff autoprogram documentation sphinx to add
usage sections for each command to the beginning of the stestr manual.
This is a standard thing to list out each option in the cli which we
were missing from the docs. Luckily cliff provides an interface to
generate this so we don't have to write it out manually.

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -3,6 +3,15 @@
 stestr user manual
 ==================
 
+Usage
+-----
+
+.. autoprogram-cliff:: stestr.cli.StestrCLI
+   :application: stestr
+
+.. autoprogram-cliff:: stestr.cm
+   :application: stestr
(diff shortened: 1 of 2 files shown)


mtreinish:
hmm, it looks like we don't publish doc updates anywhere in ci results. We do test they build properly in travis, but I only see the log there. I'll look into fixing that in a separate commit.

mtreinish:
I did take a look at this today, and travis doesn't appear to support hosting generated artifacts. It relies on pushing them to somewhere else, which makes sense. So I guess we'll just have to rely on manual local docs builds to test this.

mtreinish:
just for reference the updated docs with this patch have been uploaded: http://stestr.readthedocs.io/en/latest/MANUAL.html#usage

PR #152, opened 2018-03-09
Add an issue template to project

mtreinish:
While most of the bugs have been filed by the project maintainers it
still good to provide a template on some commonly needed information to
triage bugs. Especially including a note about --debug which is needed
to capture tracebacks.

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,18 @@
+### Issue description
+
+Include a description of the issue encountered, and what the expected behavior
+was. Also include any configuration, test suite, or other relevant information
+to reproduce the issue.
+
+Note if you encounter an error message try to re-run the command with --debug to print any tracebacks
+
+### System information
+
+**stestr version (`stestr --version`):**
+
+**Python release (`python --version`):**
+
+**pip packages (`pip freeze`):**
+
+
+### Additional information


mtreinish:
I'm not sure on the wording, but this is a start. I assume this will work per: https://help.github.com/articles/creating-an-issue-template-for-your-repository/

.github/issue_template.md
```diff
@@ -0,0 +1,18 @@
+### Issue description
+
+Include a description of the issue encountered, and what the expected behavior
+was. Also include any configuration, test suite, or other relevant information
+to reproduce the issue.
```
masayukig: I think this is a very good to starting point. But I feel it is better to have description items as bullet items like below.
What do you think?

```
* Expected behavior and actual behavior
* Steps to reproduce the problem
* Specifications like the version of the project, operating system, or hardware
```

masayukig: I really love this patch idea!

mtreinish: I'm fine with bullet points, I'm not tied to the formatting in the current patch at all. My primary motivation was just to make sure we mentioned - - debug, because I always forget about that :)

masayukig: heh, make sense :)

OK, I'm going to approve this, and then, I'll push a patch for the bullet points.

PR #155, opened 2018-03-15
Fixed manual page description 

chkumar246:


stestr/cli.py
```diff
@@ -14,16 +14,18 @@
 
 from cliff import app
 from cliff import commandmanager
+from pbr import util
 from stestr import version
 
 __version__ = version.version_info.version_string_with_vcs()
+__description__ = util.cfg_to_args()['description']
```
mtreinish: Will this work? Looking at the code in https://github.com/openstack-dev/pbr/blob/master/pbr/util.py#L195 it wants the actual setup.cfg file path. The default will work if you're in the root of the repo, but what if you run stestr --help from any other directory I don't think this will work. You might be able to use pkg_resources to pull the description from the sdist after stestr is installed, but that's kinda costly and I think just hard coding the string might be easier.

chkumar246: Yup, it appears to be costly, It is also breaking the tests. I will hard code the description.

PR #158, opened 2018-04-08
Add option to suppress attachment printing on successful tests

mtreinish:
This commit adds a new cli and user_config option to suppress printing
attachments on successful test runs. Originally in subunit_trace stdout
and stderr were unconditionally printed on test success because the
theory was this was expected to be inadvertent and likely a bug so end
users would want to know if their tests were writing to to either
stream. However, this was a faulty assumption in that some test suites
write to either during normal operation and writing the contents on test
success is just polluting the output of a test run. Therefore this adds
an option to let the user configure this functionality to address that
use case.

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -385,16 +385,19 @@ that changes the default on all available options in the config file is::
       color: True
       abbreviate: True
       slowest: True
+      suppress-attachments: True
     failing:
       list: True
     last:
       no-subunit-trace: True
       color: True
+      suppress-attachments: True
     load:
       force-init: True
       subunit-trace: True
       color: True
       abbreviate: True
+      suppress-attachments: True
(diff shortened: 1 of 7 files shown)


mtreinish:
@emonty  is this is what you were looking for?

stmcginnis:
Is there a way to set this in .stestr.conf so the default is to suppress?

Personally, I would rather have things suppressed by default, then have the ability to include the output if/when needed, but I that would be a larger behavior change. But if there is a way to set a default for this, then I would be happy.

masayukig:
Yeah, we can use the user config file to suppress by default.

If no one has objections, I'll merge this soon (within a day or two).

stmcginnis:
Thanks for the info @mtreinish.

Tested locally and works for me. I think it would be good to merge.

PR #161, opened 2018-04-11
Add docs and helper script for building man page

mtreinish:
Since stestr is a cli tool having a proper man page for it is useful. We
recently spent some time cleaning up our sphinx usage to enable
generating a proper man page by default with the sphinx man builder.
However, this wasn't documented anywhere or clearly explained. This
commit attempts to address this by first adding a small helper script to
generate a man page using sphinx and then also documenting it in the
readme. Hopefully this will encourage packagers to include a man page in
packages for the next stestr release.


mtreinish:
hmm, it looks like there something weird going on with the file access on windows in one test: https://ci.appveyor.com/project/mtreinish/stestr/build/1.0.478/job/k9bm98380del4g2x#L1250 

The interesting thing there is that the rmtree call that's failing happens in addCleanup(), so there shouldn't be anything unique about that test. My only thought right now is that the subprocess being used for the test is still accessing a file in that directory. But, if that was the case the test shouldn't have exited yet and the addCleanup shouldn't be running. So I'm not sure, I'll have to dig into it later.

I wish there was a recheck command for appveyor...

masayukig:
hmm, I have no idea.. :( I don't have a Windows box either. So, I can't check it so far.. I was thinking there's a leaking of close() somewhere. But if so, errors might be happened more (almost every time)..

README.rst
```diff
@@ -100,3 +100,18 @@ this from your project repo just call::
     $STESTR_SOURCE_DIR/tools/testr_to_stestr.py
 
 and you'll have a ``.stestr.conf`` created.
+
+Building a manpage
+------------------
+
+The stestr manual has been formatted so that it renders well as html and as a
+manpage. The html output and is autogenerated and published to:
```
masayukig: "The html output is autogenerated"?

masayukig:
@mtreinish, thanks!! I don't think the error was caused by this patch. So, I'll merge this.

PR #168, opened 2018-05-07
Fix abbreviate option in run subcommand

masayukig:
This commit fixes to be able to specify `abbreviate` option from not
only a user config file but also a command line. We should be able to
switch the flag from a command line.


mtreinish:
LGTM, nice catch. I was going to say I used abbreviate the other day, but I don't have a user config file on my system. This was a bug only if you were specifying the arg and had a user config file.

PR #171, opened 2018-05-23
Remove Python 2.6 code

masayukig:
This commit removes the code for Python 2.6 version. We don't test
stestr on Python 2.6 and we also declare the versions only 3.6, 3.5
and 2.7 in setup.cfg. So, we don't need to have this code anymore.


mtreinish:
Looks like this dates back to the original fork from testr. It looks like I just copy and pasted it from testr directly and don't remember doing that consciously. It passes CI, so it's probably safe todo.

PR #172, opened 2018-05-30
Clarify docs on group_regex

zaneb:
Interpreting the manual section about Grouping Tests required
remembering that match() only matches at the start of a string; and that
group(0) is the entire matching portion of the regex, not the first
group. State this explicitly instead.


mtreinish:
Heh, I thought I was being explicit by referencing the exact call being nsde. But I can see how that would be confusing if someone is not intimately familiar with the stdlib re moduoe. This hopefully will make things clearer for people.

PR #174, opened 2018-05-31
Really uncap pbr

mtreinish:
The previous commit attempted to uncap pbr to see if the windows issue
was addressed. But neglected to do so in the setup.py cap in addition to
the requirements file. This commit corrects the oversight and uncaps
that too.


mtreinish:
This is a release blocker for 2.1.0, once we're able to uncap pbr we'll push the new release

PR #178, opened 2018-06-07
Cleanup argument help text on load command

mtreinish:
During the migration to cliff for the 2.0.0 release we accidentally
swapped the description for some arguments to the load command. This
commit corrects this and corrects the help text for all the arguments.


mtreinish:
@masayukig it loos like this was just a rebase issue, but I'm not sure if the wording for the files argument has enough detail? What are your thoughts? If you're ok with it feel free to just merge this.

masayukig:
Yeah, I think it's enough. If we need more details, we can add it with a following patch.

PR #179, opened 2018-06-14
Cleanup the manpage section on dealing with failed tests

mtreinish:
This commit address some issues in the docs section about dealing with
failed tests. There were some formatting issues, as well as just factual
mistakes on how the commands work. This tries to address those to make
it easier to read.


mtreinish:
Hmm, yeah we build the docs in travis, but it doesn't look like it uploads the output anywhere. That's probably something we should fix. I wonder what are artifact space limit is though

Issue #181, opened 2018-06-15
bad character range while listing or running refstack tests


chkumar246:
### Issue description
Download the refstack tests:
```
wget "https://refstack.openstack.org/api/v1/guidelines/2018.02/tests?target=platform&type=required&alias=true&flag=false" -O 2018.02-test-list.txt
```

then list or run refstack tests using stestr:

```
(.venv) [chkumar246@fedora stestr]$ stestr list --whitelist-file 2018.02-test-list.txt 
bad character range
(.venv) [chkumar246@fedora stestr]$ stestr list --whitelist-file 2018.02-test-list.txt --debug
bad character range
Traceback (most recent call last):
  File "/home/chkumar246/arena/stackers/stestr/.venv/lib/python2.7/site-packages/cliff/app.py", line 399, in run_subcommand
    result = cmd.run(parsed_args)
  File "/home/chkumar246/arena/stackers/stestr/.venv/lib/python2.7/site-packages/cliff/command.py", line 184, in run
    return_code = self.take_action(parsed_args) or 0
  File "/home/chkumar246/arena/stackers/stestr/stestr/commands/list.py", line 72, in take_action
    filters=filters)
  File "/home/chkumar246/arena/stackers/stestr/stestr/commands/list.py", line 121, in list_command
    cmd.setUp()
  File "/home/chkumar246/arena/stackers/stestr/stestr/test_processor.py", line 148, in setUp
    black_regex=self.black_regex)
  File "/home/chkumar246/arena/stackers/stestr/stestr/selection.py", line 113, in construct_list
    list_of_test_cases = filter_tests(regexes, test_ids)
  File "/home/chkumar246/arena/stackers/stestr/stestr/selection.py", line 29, in filter_tests
    _filters = list(map(re.compile, filters))
  File "/home/chkumar246/arena/stackers/stestr/.venv/lib64/python2.7/re.py", line 194, in compile
    return _compile(pattern, flags)
  File "/home/chkumar246/arena/stackers/stestr/.venv/lib64/python2.7/re.py", line 251, in _compile
    raise error, v # invalid expression
error: bad character range
Traceback (most recent call last):
  File "/home/chkumar246/arena/stackers/stestr/.venv/bin/stestr", line 10, in <module>
    sys.exit(main())
  File "/home/chkumar246/arena/stackers/stestr/stestr/cli.py", line 101, in main
    return cli.run(argv)
[...]

PR #182, opened 2018-06-15
Add error handling for invalid input regexes

mtreinish:
This commit adds error handling for invalid input regexes on any of the
test selection mechanisms. Now when an invalid regex is encountered an
error message will be printed and then stestr will exit. This should hopefully
make it easier to identify bad user input, like described in #181.

The logic in selection.py needed to be reworked slightly to
differentiate between invalid regex in input filters and the whitelist
file.

### DIFF
--- a/stestr/selection.py
+++ b/stestr/selection.py
@@ -12,6 +12,7 @@
 
 import contextlib
 import re
+import sys
 
 
 def filter_tests(filters, test_ids):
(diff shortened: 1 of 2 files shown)


stestr/selection.py
```diff
@@ -26,7 +27,17 @@ def filter_tests(filters, test_ids):
...
         return test_ids
-    _filters = list(map(re.compile, filters))
+
+    _filters = []
+    for f in filters:
+        if isinstance(f, str):
+            try:
+                _filters.append(re.compile(f))
+            except re.error:
+                print("Invalid regex: %s provided in filters" % f)
```
masayukig: I was thinking like these error messages should be written to stderr. But it's a separate thing. We already do it in the others.
I'll push a patch to fix it if we should. What do you think?

mtreinish: Yeah using stderr is probably the proper way to handle this. It's a pattern I've used in other places, mostly because I'm lazy. I think you pushing a follow up patch to switch all of those places to use stderr is a good idea.

masayukig: OK, cool. I'll push a patch for that later.

PR #183, opened 2018-06-18
Make warning and error messages use stderr

masayukig:
This commit makes warning and error messages use stderr for
output. Originally, those messages use stdout. However, those messages
should use stderr instead because of the responsibility. This commit
keeps the original exit codes for the backward compatibility.

Related-PR: #182


ssbarnea:
I support use of stderr, as this is the standard (posix) stream to use for any error/warning/info/debug messages.

stestr/subunit_trace.py
```diff
@@ -386,7 +387,7 @@ def trace(stdin, stdout, print_failures=False, failonly=False,
     # NOTE(mtreinish): Ideally this should live in testtools streamSummary
     # this is just in place until the behavior lands there (if it ever does)
     if count_tests('status', '^success$') == 0:
-        print("\nNo tests were successful during the run")
+        print("\nNo tests were successful during the run", file=sys.stderr)
```
mtreinish: This one is even more on the edge for me. I'm not sure if stderr is better here or not. In this case we ran the tests and there were tests being run, just none were successful. Normally this gets printed if you run a test suite and all the cases in the suite are skipped. Which in the normal unittest/testtools world would be a successful run. We handle this separately here because that's normally not the desired result from a test run to skip all the tests and normally indicates a failure. All the other subunit-trace output in this case goes to stdout, I'm not sure this is sufficiently different to warrant being in stderr.

stestr/subunit_trace.py
```diff
@@ -376,7 +377,7 @@ def trace(stdin, stdout, print_failures=False, failonly=False,
     elapsed_time = stop_time - start_time
 
     if count_tests('status', '.*') == 0:
-        print("The test run didn't actually run any tests")
+        print("The test run didn't actually run any tests", file=sys.stderr)
```
mtreinish: This is the only place I'm unsure of. (and below) The way this is handled it's the normal flow for subunit-trace and ran through the stream and no test cases were encountered. I'm not sure if this should go to stderr since normal subunit-trace output all goes to stdout.

ssbarnea:
My personal preference is to use logging for messages as that goes directly to stderr and is easier to tune. But, even print with stderr is ok. At least the messages from this PR seem to be aimed for stderr.

When deciding what goes to stdout vs stderr, I think you always have to consider that stdout is the "parsable" output of the program, when stderr is the non-parsable. For unix tools like find/grep/ack/ls is easy to think. For example ls will only output filenames on stdout (one on each line when piped), all other errors/warnings will go to stderr, for obvious reasons. 

Having this separation is key as it assures that a consumer could parse the stdout. And yes, is perfectly normal to have no stdout output, even when you get an error. Message would go to stderr and the application would return an error code.

Based on my experience with CI and testing tools, I found one particular case where bugs slept in: when no tests were found/executed and the tool returned an empty set of tests with 0 exit code. The hack here is to add an extra option to tell, fail-on-no-tests, maybe even better: a numeric value: min-tests-to-run, which would return an non zero error code when total number of tests executed was less than that value. Having something like this would make everyone happy.

At this moment we had to implement few safety checks where we decide to fail a build if it produces no  tests.

mtreinish:
@masayukig ok that makes sense. I think this is fine now. I'm gonna go ahead and merge it.

@ssbarnea if you want to switch the print() calls to use logging that'd probably be a welcome patch.

PR #187, opened 2018-07-10
WIP: Fix unxsuccess return code

mtreinish:
This is a first attempt to fix unxsuccess. It doesn't quite work, my
guess is because likely the test result is properly tracking the
unxsuccess tests and thus the list is empty when it's manually checked.


mtreinish:
Just for reference this is the same issue as: https://bugs.launchpad.net/testrepository/+bug/1429196 it's definitely a deeper issue, and this quick fix won't work

PR #188, opened 2018-07-10
WIP: Fix unxsuccess return code

mtreinish:
This is commit fixes the response code from runs using subunit trace
with unxsuccess test results. Currently the test result object isn't
properly tracking the unxsuccess tests, so when we check if the run was
successful when it looks for any unxsuccess tests it doesn't see any and
returns 0. Since this is happening in a lower level for the time being
this fixes it at a higher level using subunit_trace to manually count the
unxsuccess tests. This however doesn't fix the case of
--no-subunit-trace, which will still not work because it never calls the
subunit_trace module.


cdent:
This seems a reasonable workaround for now, but does seem a bit...kludgy?

mtreinish:
Yeah, it's definitely not a pretty solution. But it does seem to work. I'm thinking I maybe should break out the counter (which just leverages testtools.StreamToDict() function into a separate helper so we can use the same basic method of working around the underlying issue outside of subunit_trace. 

But before that, I still have to figure out what this breaks in the tests.

PR #190, opened 2018-08-03
Fix handling of unexpected success results

### LINKED ISSUE #189: Unexpected Successes do not result in overall failure
### Issue description

It appears that tests that pass unexpectedly, when marked with `@unittest.expectedFailure` do not trigger an overall test suite failure.

Based on changes in python3.4 from the fix landing for https://bugs.python.org/issue20165 this is rather unexpected.

Mark any successful test with the decorator `@unittest.expectedFailure`
Run the tests using `stestr run --slowest`
Exit code should be non-zero

Actual behaviour is that current exit code is zero (0)

Mark any successful test with the decorator `@unittest.expectedFailure` and run tests with stestr.

### System information

**stestr version (`stestr --version`):** stestr 2.1.0

**Python release (`python --version`):** Python 3.5.2

**pip packages (`pip freeze`):**
-f /var/cache/pip
cliff==2.13.0
cmd2==0.9.2
colorama==0.3.9
extras==1.0.0
fixtures==3.0.0
flake8==3.5.0
future==0.16.0
linecache2==1.0.0
<have to prune internal project from this list>
mccabe==0.6.1
pbr==4.1.0
pkg-resources==0.0.0
prettytable==0.7.2
pycodestyle==2.3.1
pyflakes==1.6.0
pyparsing==2.2.0
pyperclip==1.6.2
python-mimeparse==1.6.0
python-subunit==1.3.0
PyYAML==3.13
six==1.11.0
stestr==2.1.0
stevedore==1.28.0
testtools==2.3.0
traceback2==1.4.0
unittest2==1.1.0
voluptuous==0.11.1
wcwidth==0.1.7

### Additional information

It would appear that testtools TestResult class contains a method that has the correct behaviour at https://github.com/testing-cabal/testtools/blob/master/testtools/testresult/real.py#L174 and it would be possible to fix this in stestr by changing from calling `summary_result.wasSuccessful()` at https://github.com/mtreinish/stestr/blob/master/stestr/commands/load.py#L222 to call `testtools.TestResult.wasSuccessful(summary_result)` instead to resolve. There would also be a few other places that would need updating.

Based on looking at the tests in testing-cabal/testtools It may be that testtools is maintaining backwards compatibility with `unittest.TestResult` prior to the fix landing in python 3.4 in the StreamSummary class and downstream projects at expected to handle this instead. Will log an issue there as well referencing this issue to try and determine where is the correct place to submit a fix for this.

mtreinish:
This commit addresses a bug when unexpected success results are used
by a test suite. stestr was relying on wasSuccessful() method in the
StreamResult class from testtools, which doesn't properly account for
unexpected success. So when stestr was run on a test suite it wouldn't
properly handle tests with unexpected success result and thus treat
the run as successful. This addresses the issue by adjusting our
testtools api usage to call the wasSuccessful() method from the
TestResult class instead which has the correct behavior.

An issue has been opened in testtools testing-cabal/testtools#273
regarding the accounting error with StreamResult.wasSuccessful(),
if/when a fix is introduced into a released testtools we can investigate
switching these calls back to use that function. But for the time being
this mitigates the issue.

Fixes #189

### DIFF
--- a/stestr/commands/failing.py
+++ b/stestr/commands/failing.py
@@ -113,7 +113,7 @@ def failing(repo_type='file', repo_url=None, list_tests=False, subunit=False,
         case.run(result)
     finally:
         result.stopTestRun()
-    failed = not summary.wasSuccessful()
+    failed = not results.wasSuccessful(summary)
     if failed:
         result = 1
     else:
--- a/stestr/commands/last.py
+++ b/stestr/commands/last.py
@@ -154,7 +154,7 @@ def last(repo_type='file', repo_url=None, subunit_out=False, pretty_out=True,
             case.run(output_result)
         finally:
             output_result.stopTestRun()
-        failed = not summary.wasSuccessful()
+        failed = not results.wasSuccessful(summary)
     else:
         stream = latest_run.get_subunit_stream()
         failed = subunit_trace.trace(stream, stdout, post_fails=True,
--- a/stestr/commands/load.py
+++ b/stestr/commands/load.py
@@ -219,7 +219,7 @@ def load(force_init=False, in_streams=None,
     if pretty_out and not subunit_out:
         subunit_trace.print_fails(stdout)
         subunit_trace.print_summary(stdout, elapsed_time)
-    if not summary_result.wasSuccessful():
+    if not results.wasSuccessful(summary_result):
         return 1
     else:
         return 0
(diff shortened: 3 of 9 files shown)


mtreinish:
Ugh, python2 is enforcing the object's type for the self argument. While python3 will take a different type we pass in. We'll have to figure out a different method of doing this I guess, maybe just building a helper to do manually what testtools.TestResult.wasSuccessful() is doing.

mtreinish:
Heh, because I missed it. I'll push up a new revision in a minute to correct the oversight.

mtreinish:
@masayukig  ok, the latest revision now has the last command updated too

PR #191, opened 2018-08-07
Add support for python 3.7

mtreinish:
Python 3.7 is out and some distros are already using it (including mine,
archlinux) so we should add official support for it. It doesn't appear
that there are any functional changes needed. This commit just updates
the package metadata and adds python 3.7 test jobs.


mtreinish:
hmm, I guess we'll need to figure out how to change the distro it's running on for travis. Because the trusty image it's defaulting to won't work with >=3.7 because of a openssl lib. (which is why the binary to download/run isn't found during the job)

mtreinish:
@masayukig do you have any idea what I'm doing wrong in the travis config? According to every reference I've found by searching says that you just specify `dist: xenial` and `python: 3.7` and it should work.

mtreinish:
This is the last patch in the queue after we figure this out and merge it I'm planning to push 2.1.1 so we can get the unxsuccess patch out there for people.

masayukig:
I'm not sure but this could help?

https://github.com/travis-ci/travis-ci/issues/9815#issuecomment-401756442
```
language: python
python: 3.7
dist: xenial
sudo: true
```
add `sudo: true` ?

mtreinish:
@masayukig hah, well it looks like `sudo: true` was the missing piece after all.

masayukig:
awesome :)

PR #192, opened 2018-09-06
Add support for test class and method by path on no-discover

### LINKED ISSUE #180: [RFE] pytest-style filesystem paths
The pytest runner allows you to provide a filepath-like arguments to run tests instead of paths runners.

    pytest test_mod.py::TestClass::test_method

We already have support for the path aspect of this, using the `--no-discover`/`-n` option but this does not allow you specify a class/method.

    stestr -n test_mod.py

To do this, you need to use a Python path.

    stestr test_mod.TestClass.test_method

It would be helpful to support the pytest-style case as this is frequently quicker to run and doesn't require discovery outside of the file provided (or so I assume).

mtreinish:
This commit adds support to the --no-discover/-n flag on stestr run for
specifying a test class and test method when passing in a filepath. This
borrows the syntax from pytest so you use '::' to specify a class and
then method. For example 'path/test_mod.py::TestClassA::test_method_b'
would be used to run only test_method_b in the TestClassA test case in
the path/test_mod.py file.

Fixes #180

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -313,8 +313,10 @@ def run_command(config='.stestr.conf', repo_type='file',
         combine_id = six.text_type(latest_id)
     if no_discover:
         ids = no_discover
+        if '::' in ids:
+            ids = ids.replace('::', '.')
         if ids.find('/') != -1:
-            root, _ = os.path.splitext(ids)
+            root = ids.replace('.py', '')
             ids = root.replace('/', '.')
         run_cmd = 'python -m subunit.run ' + ids
 
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -100,6 +100,14 @@ import paths) For example::
 will also bypass discovery and directly call subunit.run on the module
 specified.
 
+Additionally you can specify a specific class or method within that file using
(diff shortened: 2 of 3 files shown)


mtreinish:
@stephenfin let me know if this works for you

stestr/commands/run.py
```diff
@@ -313,10 +313,13 @@ def run_command(config='.stestr.conf', repo_type='file',
...
     if no_discover:
         ids = no_discover
+        if '::' in ids:
+            ids = ids.replace('::', '.')
         if ids.find('/') != -1:
-            root, _ = os.path.splitext(ids)
+            root = ids.replace('.py', '')
             ids = root.replace('/', '.')
         run_cmd = 'python -m subunit.run ' + ids
+        print(run_cmd)
```
masayukig: Should this be removed?

mtreinish: Oops, yeah this was leftover from my from my debugging. I'll respin it right now

masayukig:
thanks, LGTM

stephenfin:
Just because it's good to tell people this stuff, I use this religiously nowadays as part of my work on nova. Much appreciated :+1:

Issue #137, opened 2018-01-15
"WARNING: missing Worker N! Race in testr accounting." when using stestr --failing


cdent:
While iteratively using the `--failing` argument to find and fix problems in a large suite of tests (the OpenStack nova unit tests) I received 

```
WARNING: missing Worker 9! Race in testr accounting.
```

warnings for one or more workers, when spreading tests across 16 cores. As the number of tests being run shrank, the chance for the warning seemed to increase.

It seemed to be the case that a worker was reporting the problem when rationalizing the discovered tests against the list of failing tests and resulting in none.

mtreinish:
Related, but not at all relevant to fixing this bug. I just pushed up https://github.com/mtreinish/stestr/pull/138 to remove the second half of the error message. Just putting that here for the record so in case people hit this after that patch lands to indicate the second half of the message will go away.

PR #193, opened 2018-10-03
Use reported times instead of wall time in subunit-trace

### LINKED ISSUE #119: stestr load run time output doesn't reflect reality
When you run stestr load on a stream not being processed in real time with stestr run (or via stdin with a subunit emitting test run)  the run time included in the output generated via the subunit_trace module doesn't reflect reality. Instead it is the time it took to process the subunit stream which isn't accurate. We should update the subunit trace module to figure out the run time from the timestamps inside the stream instead of taking 2 timestamps (one before the other after) around the subunit processing to figure out the run time.

mtreinish:
This commit updates the subunit_trace module to leverage the timestamps
reported in the subunit instead of the wall time for the elapsed time
output. When we were using the wall time if we were just reading a
subunit file this would be the time it took to read an process the
subunit data. Which is not the actual run time of the test. By switching
to use the elapsed time between the first start timestamp and the last
stop timestamp we'll always get an accurate value. This does come with
the tradeoff of not counting any setUp before the start time or
tearDown/cleanUp after the last test, since that data isn't in the
subunit stream. But that feels like a worthwhile compromise for having
moderately accurate numbers all the time.

Fixes #119

### DIFF
--- a/stestr/commands/load.py
+++ b/stestr/commands/load.py
@@ -13,7 +13,6 @@
 """Load data into a repository."""
 
 
-import datetime
 import functools
 import os
 import sys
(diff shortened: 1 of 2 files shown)


mtreinish:
We should test that this actually fixes #119 before we merge this

mtreinish:
Yeah I got that too on my environment, but I don't think it's being caused by this patch. I'm seeing the same thing on the last release (2.1.1) too. There is an open bug related to this #137 that's about stest run --failing, but I suspect this is the same underlying cause (but I have no idea what that is). Especially because that will call the load command internally too.

masayukig:
OK, yeah, absolutely.

PR #195, opened 2018-10-05
Make test running serially when just loading

### LINKED ISSUE #137: "WARNING: missing Worker N! Race in testr accounting." when using stestr --failing

masayukig:
This commit tries to fix the issue #137 "WARNING: missing Worker N!"
when using `stsetr run --failing` / `stestr load ..`.

We shouldn't use `testtools.ConcurrentStreamTestSuite` when we load
past subunit data. Because it doesn't match the load workers against
the run workers.

Fixes #137

### DIFF
--- a/stestr/commands/load.py
+++ b/stestr/commands/load.py
@@ -106,13 +106,14 @@ class Load(command.Command):
              force_init=force_init, streams=args.files,
              pretty_out=pretty_out, color=color,
              stdout=stdout, abbreviate=abbreviate,
-             suppress_attachments=suppress_attachments)
+             suppress_attachments=suppress_attachments, serial=True)
 
 
 def load(force_init=False, in_streams=None,
          partial=False, subunit_out=False, repo_type='file', repo_url=None,
          run_id=None, streams=None, pretty_out=False, color=False,
-         stdout=sys.stdout, abbreviate=False, suppress_attachments=False):
+         stdout=sys.stdout, abbreviate=False, suppress_attachments=False,
+         serial=False):
     """Load subunit streams into a repository
 
     This function will load subunit streams into the repository. It will
@@ -180,7 +181,13 @@ def load(force_init=False, in_streams=None,
             case = testtools.DecorateTestCaseResult(case, decorate)
             yield (case, str(pos))
 
-    case = testtools.ConcurrentStreamTestSuite(make_tests)
+    if serial:
+        for _, stream in enumerate(streams):
+            # Calls StreamResult API.
+            case = subunit.ByteStreamToStreamResult(
+                stream, non_subunit_name='stdout')
+    else:
+        case = testtools.ConcurrentStreamTestSuite(make_tests)
     if not run_id:
         inserter = repo.get_inserter()
     else:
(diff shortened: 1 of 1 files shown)


stestr/commands/load.py
```diff
@@ -180,7 +181,13 @@ def make_tests():
             case = testtools.DecorateTestCaseResult(case, decorate)
             yield (case, str(pos))
 
-    case = testtools.ConcurrentStreamTestSuite(make_tests)
+    if serial:
+        for _, stream in enumerate(streams):
```
mtreinish: Is there a reason we need to use enumerate here instead of just doing `for stream in streams:` ?

stestr/commands/load.py
```diff
@@ -180,7 +181,13 @@ def make_tests():
...
+            # Calls StreamResult API.
+            case = subunit.ByteStreamToStreamResult(
+                stream, non_subunit_name='stdout')
```
mtreinish: What happens if there is more than one stream we tell it to load? Doesn't this ignore all but the last stream?

mtreinish: I'm thinking to get around that we probably should make like L195-L238 a separate function like `_load_case()` or something. Then inside the loop here and after L190 call it with the ConcurrentStreamTestSuite case we call that.

stestr/commands/load.py
masayukig: ah, yes. I should work. I just copy the code from `make_tests()` and modified it :-p

stestr/commands/load.py
masayukig: for sure, I'll do that.

masayukig: well, I'm a bit confused. If we separate like that, what about L191-L194? I probably misunderstand something, though.. I guess we should make L19**1**-L238 a separate function to `_load_case()`?

mtreinish: For this approach to to work we need to have L191-194 outside the loop (well really just L192 since that creates a new run in the repository, but they kind of all go together) Since the new function will be called from inside the loop we can't have the inserter be allocated inside the function. We'll have to move L191-194 above the loop and then pass the inserter object into `_load_case()`

masayukig:
Like this?

mtreinish:
Yep, that lgtm it should address my earlier concerns with ignoring all but the last stream in the earlier revision. My only question before I approve this is on behavior now though. So with the way retval is handled now if 1 stream has a failure in it we stop loading the other streams passed in. Not sure if that's a problem but it is a change in behavior from what we have now (or at least I think is). Wouldn't we want all the streams added to the repository even if some have test failures?

masayukig:
oh, I was thinking I wanted to make it like that. I mean we want all the streams added to the repository even if some have test failures. I'll update my PR soon.

Issue #97, opened 2017-09-19
Integrate with coverage


pshchelo:
To be a more complete replacement for 'testrepository', stestr  should support setuptools integration (with coverage integration) as testrepository does 

http://testrepository.readthedocs.io/en/latest/MANUAL.html#setuptools-integration

mtreinish:
I'm not sure if we want to do this. I've always viewed the testr setuptools integration (either exposed directly by testr or via pbr) as a problem because it makes things more confusing for end users. People don't really end up knowing where and how their tests are being run when you hide the test runner execution behind an external tool. (for a recent example: https://github.com/testing-cabal/testrepository/issues/31 ) IMO it would be better to just add any functionality to the stestr command and document how to use it.

It sounds like what you really want is a coverage flag for stestr so users don't have to explicitly set PYTHON=coverage run --parallel. I don't think we need to add a distutils.commands entrypoint just to do that though. Is there anything else you feel the setuptools integration buys us?

pshchelo:
@mtreinish not really, coverage integration would be enough. Overriding PYTHON does seems weird, although doable using tox->stestr.

I'll change the issue title accordingly.

mtreinish:
Well if you want to just do it in tox it's not that bad. For example what I did in shade:

https://github.com/openstack-infra/shade/blob/master/tox.ini#L47-L55

But I think integrating a coverage option into stestr is a good feature to add. Not sure when I can get around to implementing it myself, but feel free to push patches if you'd like to give it a try.

masayukig:
Just FYI, I started to implement for this issue on my branch https://github.com/masayukig/stestr/tree/integrate-coverage

masayukig:
@mtreinish heh, thanks :)

masayukig:
@mtreinish So, my current implementation[0] is on #100 patch because I was expecting the cliff patch would be merged soon. But if you want this feature earlier, I'll change the patch.
Thoughts?

masayukig:
Well, we decided to close this issue with this conversation at this moment. But if we change our minds again in the future, we can always reopen it. Any comments are welcome anytime.

https://github.com/mtreinish/stestr/pull/196#pullrequestreview-163859660
> That being said I have been thinking about this some more and wonder if this goes against the unix philosophy for stestr. stestr is a parallel python test runner and it should do that well, while coverage collection and reporting is the role of the coverage tool.

PR #196, opened 2018-10-10
Integrate with coverage

masayukig:
This commit adds `--coverage` option to integrate with `coverage`.

Fixes Issue #97

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -131,6 +131,10 @@ class Run(command.Command):
                             help='If set do not print stdout or stderr '
                             'attachment contents on a successful test '
                             'execution')
+        parser.add_argument('--coverage', action='store_true',
+                            dest='coverage',
+                            help='Replace PYTHON with coverage and merge '
+                            'coverage from each testr worker.')
         return parser
 
     def get_description(self):
@@ -187,7 +191,7 @@ class Run(command.Command):
             combine=args.combine,
             filters=filters, pretty_out=pretty_out, color=color,
             stdout=stdout, abbreviate=abbreviate,
-            suppress_attachments=suppress_attachments)
+            suppress_attachments=suppress_attachments, coverage=args.coverage)
 
         # Always output slowest test info if requested, regardless of other
         # test run options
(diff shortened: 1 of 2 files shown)


stestr/commands/run.py
```diff
@@ -318,19 +334,32 @@ def run_command(config='.stestr.conf', repo_type='file',
...
+                               subunit_out=subunit_out,
+                               repo_type=repo_type,
+                               repo_url=repo_url, run_id=combine_id,
+                               pretty_out=pretty_out,
+                               color=color, stdout=stdout,
+                               abbreviate=abbreviate,
+                               suppress_attachments=suppress_attachments)
+            if coverage:
+                os.system("coverage combine")
+                os.system("coverage html -d ./cover %s" % omit)
```
mtreinish: I don't think that we want to generate html output like this from a CLI tool. If we're going to generate the report (and not just collect coverage)  I think we should call coverage report to print to stdout instead.

stestr/commands/run.py
```diff
@@ -318,19 +334,32 @@ def run_command(config='.stestr.conf', repo_type='file',
...
+            result = load.load(in_streams=run_proc,
+                               subunit_out=subunit_out,
+                               repo_type=repo_type,
+                               repo_url=repo_url, run_id=combine_id,
+                               pretty_out=pretty_out,
+                               color=color, stdout=stdout,
+                               abbreviate=abbreviate,
+                               suppress_attachments=suppress_attachments)
+            if coverage:
+                os.system("coverage combine")
```
mtreinish: Is there a python interface/api we could call instead of shelling out?

masayukig:
> That being said I have been thinking about this some more and wonder if this goes against the unix philosophy for stestr. stestr is a parallel python test runner and it should do that well, while coverage collection and reporting is the role of the coverage tool.

Yeah, good point.
After reading your comment, I think we don't need this function so far because this PR brings us little benefit.

I was thinking this function could be useful a bit, especially for original `.... testr --coverage` users. But, now, I think the benefit is not that much. And, yea, we can do same things with the coverage tool. Another concern is that this could introduce another complexity to stestr itself..

So, let's close this PR and the issue #97 ?

mtreinish:
Well, I'm not sure. At one point (mainly back when #97 was opened) I felt like it was in scope for stestr. But, now if feels out of scope to me. So I guess we should close it and if we change our minds again in the future :) we can always reopen it.

masayukig:
yeah, agree. Let's close both of the PR and issue at this moment.

PR #197, opened 2018-10-11
Enable doc8

masayukig:
This commit enables doc8 checks to validate rst files.


masayukig:
heh, yeah. We've done for the documents very well so far.

PR #200, opened 2018-10-22
Introduce parallel-class option

### LINKED ISSUE #198: Add config/cli flag to group tests by class
### Issue description

Currently we only support dynamic grouping (ignoring manual scheduling) via a grouping regex. This is fine and very flexible, but it's tricky to construct a regex that works exactly as needed. A common use case is that a test suite needs to be parallelize on the test class and not individual methods/test_ids. This is currently accomplishable by just setting the group_regex to: `([^\.]*\.)*`. But, it would probably be easier/better to add a new option to the stestr.conf and cli to split at the class level so people don't need to remember that regex.

I'm just not sure of the best name for that --parallel-class or something?

Also something we'll have to put some thought and discussion behind is the priority of arguments. Like does --parallel-class (or whatever we call it) take priority over group_regex. Also between parallel-class in a config file and --group-regex on the cli (and vice versa)

masayukig:
This commit introduces the parallel-class option to the CLI and config
file. The option take priority over group_regex. And if both on the
CLI and in the config are set, we use the option on the CLI not in a
config file.

For example, `--group-regex` on the CLI and `parallel-class` in a
config file are set, `--group-regex` is higer priority than
`parallel-class` in this case.

Fixes #198

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -356,7 +356,11 @@ method)::
 
     group_regex=([^\.]+\.)+
 
-Or, you can use the following option instead::
+However, because grouping tests at the class level is a common use case there is also a config option, ``parallel_class``, to do this. For example, you can use::
+
+    parallel_class=True
+    
+and it will group tests in the same class together.
 
     parallel_class=True
 
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -74,9 +74,10 @@ The ``group_regex`` option is used to specify is used to provide a scheduler
 hint for how tests should be divided between test runners. See the
 :ref:`group_regex` section for more information on how this works.
 You can also specify the ``parallel_class=True`` instead of
-``group_regex=([^\.]*\.)*``. This ``parallel_class`` option is to
-group tests together in the stestr scheduler by class. So, you don't
-need to memorize the complicated regex with it.
+group_regex to group tests in the stestr scheduler together by
+class. Since this is a common use case this enables that without
+needing to memorize the complicated regex for ``group_regex`` to do
+this.
 
 There is also an option to specify all the options in the config file via the
 CLI. This way you can run stestr directly without having to write a config file
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -73,6 +73,10 @@ A full example config file is::
 The ``group_regex`` option is used to specify is used to provide a scheduler
 hint for how tests should be divided between test runners. See the
 :ref:`group_regex` section for more information on how this works.
+You can also specify the ``parallel_class=True`` instead of
+``group_regex=([^\.]*\.)*``. This ``parallel_class`` option is to
+group tests together in the stestr scheduler by class. So, you don't
+need to memorize the complicated regex with it.
 
 There is also an option to specify all the options in the config file via the
 CLI. This way you can run stestr directly without having to write a config file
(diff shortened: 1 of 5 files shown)


mtreinish:
Sigh, github doesn't seem to be working, so I can't actually view the code (it took effort just to open the PR, it kept redirecting me to issue 200 instead of pull and getting a 404) But without being able to review the code yet (assuming it's a backend sync issue on github's side that will fix itself eventually) based on the notification email this is missing documentation in the MANUAL.rst. We should document the new option, how it's used and it's relationship with group-regex.

mtreinish:
Sigh, github doesn't seem to be working, so I can't actually view the code (it took effort just to open the PR, it kept redirecting me to issue 200 instead of pull and getting a 404) But without being able to review the code yet (assuming it's a backend sync issue on github's side that will fix itself eventually) based on the notification email this is missing documentation in the MANUAL.rst. We should document the new option, how it's used and it's relationship with group-regex.

mtreinish:
Sigh, github doesn't seem to be working, so I can't actually view the code (it took effort just to open the PR, it kept redirecting me to issue 200 instead of pull and getting a 404) But without being able to review the code yet (assuming it's a backend sync issue on github's side that will fix itself eventually) based on the notification email this is missing documentation in the MANUAL.rst. We should document the new option, how it's used and it's relationship with group-regex.

masayukig:
I'm sorry for my delay. For sure, additional documentation is needed. I'll update this PR. And also, I'll revisit your comments inline later.

mtreinish:
@masayukig have you had a chance to revisit this? If not I can give making the modifications a try.

masayukig:
ah, I'm sorry for no progress for a while. I actually don't have time for now. So, I'll do this later this week. But if it's too late, could you give them?

doc/source/MANUAL.rst
```diff
@@ -352,6 +356,18 @@ method)::
 
     group_regex=([^\.]+\.)+
 
+Or, you can use the following option instead::
```
mtreinish: How about instead of this we write:


```suggestion
However, because grouping tests at the class level is a common use case there is also a config option, ``parallel_class``, to do this. For example, you can use::

    parallel_class=True
    
and it will group tests in the same class together.
```

README.rst
```diff
@@ -22,12 +22,22 @@ You can see the full rendered docs at: http://stestr.readthedocs.io/en/latest/
...
+
+.. _unittest: https://docs.python.org/3/library/unittest.html
+.. _subunit: https://github.com/testing-cabal/subunit
+
+stestr originally started as a fork of the `testrepository`_ that concentrates
+on being a dedicated test runner for python projects. The generic abstraction
+layers which enabled testrepository to work with any subunit emitting runner
+are gone. stestr hard codes python-subunit-isms into how it works. The code
+base is also designed to try and be explicit, and to provide a python api that
+is documented and has examples.
```
mtreinish: Hmm, why is this in the diff? This seems like a rebase issue.

doc/source/MANUAL.rst
```diff
@@ -73,6 +73,10 @@ A full example config file is::
 The ``group_regex`` option is used to specify is used to provide a scheduler
 hint for how tests should be divided between test runners. See the
 :ref:`group_regex` section for more information on how this works.
+You can also specify the ``parallel_class=True`` instead of
```
mtreinish: Can we make this a separate paragraph (just add a blank line). Also I think we should reorder it so we say that it groups by class before saying the equivalent regex. Basically something like:

```suggestion

You can also specify the ``parallel_class=True`` instead of group_regex to group tests in the stestr scheduler together by class. Since this is a common use case this enables that without needing to memorize the complicated regex for ``group_regex`` to do this.
```

masayukig:
oops, I'm bad at the github operation.. I should have pushed "Add suggestion to batch" instead :(

README.rst
masayukig: yeah, I'll try to fix that.

masayukig:
hmm, I'm wrong something about git rebase.. Let me try once more again..

doc/source/MANUAL.rst
masayukig: Thanks for the suggestion!

masayukig:
Now, it's cleaner than before.

PR #203, opened 2018-11-16
Fix discovery import error formatting on py3

mtreinish:
This commit fixes the formatting when we print the errors encountered
during discovery. Because python3 returns a bytes object from subprocess
just casting it as a str() won't make the ouptut respect the the
newlines and tabs. To fix the issue this commit splits out the py3 case
and properly decodes the bytes as utf8 data and pass that to
sys.stdout.write().


masayukig:
Sure. Yeah, I totally agree with you. This is annoying.

PR #205, opened 2018-11-26
Change title of project in readme

mtreinish:
No one calls stestr Slim Test Repository or Super Test Repository, so
lets stop referring to it as that. stestr should just be stestr, for
better or worse, and not stand for anything. This commit just updates
the readme to reflect this.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
-Slim/Super Test Repository
-==========================
+stestr
+======
 
 .. image:: https://img.shields.io/travis/mtreinish/stestr/master.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


mtreinish:
Heh, we can put a comment somewhere else in the docs explaining the history of the name. But I think very few people have ever asked about the name of the project. The most common question I get about the name  is because people confuse `ostestr`, `os-testr`, and `stestr`.

masayukig:
oh, I was thinking I merged this already but not yet :( I'll do that asap.
Heh, yeah, the names are confusing actually.

Issue #202, opened 2018-11-07
Add input for stestr run --concurrency flag for fractional nprocs


mtreinish:
Right now the --concurrency flag for stestr run takes in an integer, if that value is 0 it runs nproc workers, if that value is greater than 0 it runs that many workers. But we don't have a way to specify a value that is a fraction of nproc easily (like nproc/2 or nproc/4). This can be useful for some test suites, especially if they're very resource intensive, where running nproc workers isn't viable but you still want the number of workers to scale dynamically based on the size of the machine.

There are 2 ways I can see handling this first we also take a string in for concurrency and define an input for the fractions. Like `stestr run --concurrency 'n/2'` will run with nprocs / 2 workers. Or we could use the negative space and stick with integers to define the same thing, so something like `stest run --concurrency -2` will run with nprocs / 2 workers.

mtreinish:
@masayukig do you have any thoughts on this?

masayukig:
yeah, I think it's a good feature. I actually don't need like that option so far personally. But it makes sense to me.

So, regarding the option argument, I'm fine with either way. But I prefer `-2` style because it's simpler.

I was also think like `--concurrency '0.5'` but I feel that would cause a confusion and inconsistency.

mtreinish:
Well it mostly came up for me because of: https://github.com/Qiskit/qiskit-terra/pull/1228 where we're hitting occasionally memory issues. I'd rather set it to nproc/2 or nproc/4 rather than hard coding it to 2 workers.

mtreinish:
As for the specific input definition I'm fine with anything as long as we make it clear. I only suggested '-2' or 'n/2' because I thought the implementation would be simpler. If we take a float my concern is that someone will do something like `--concurrency 0.124332678212341`. Granted we will always have the cast the result as an `int` so maybe that isn't a problem.

masayukig:
Hrm. I actually don't have a strong opinion here. If we know the nproc already, the subtraction might work well. But I also think, if we know that, can we specify the concurrency directly? So, I guess we use this feature when the nproc is ambiguous. If so, I feel treating the concurrency as a fraction might be better.

And, I'm curious about a corner case of the subtraction. What will happen when nproc =< abs(concurrency) && 0 > concurrency? For example, nproc=32 and --concurrency=-32.
Of course, I think we can treat this just no concurrency in this case, though.

mtreinish:
[...]

PR #208, opened 2018-12-18
Add error message for invalid concurrency value

mtreinish:
This commit adds a detailed error message when a user passes in a
concurrency outside the expected range (which is >= 0). Previously
we'd raise an unhandled IndexError because of going outside the list
bounds deeper in the code, which wasn't super useful or informative for
users. This fixes that by detecting when we have an invalid input up
front and print that out before we run anything.

Related to: #202

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -172,6 +172,11 @@ class Run(command.Command):
             suppress_attachments = args.suppress_attachments
         verbose_level = self.app.options.verbose_level
         stdout = open(os.devnull, 'w') if verbose_level == 0 else sys.stdout
+        if concurrency and concurrency < 0:
+            msg = ("The provided concurrency value: %s is not valid. An "
+                   "integer >= 0 must be used.\n" % concurrency)
+            stdout.write(msg)
+            return 2
         result = run_command(
             config=self.app_args.config, repo_type=self.app_args.repo_type,
             repo_url=self.app_args.repo_url,
@@ -308,6 +313,11 @@ def run_command(config='.stestr.conf', repo_type='file',
             exit(1)
         repo = util.get_repo_initialise(repo_type, repo_url)
     combine_id = None
+    if concurrency and concurrency < 0:
+        msg = ("The provided concurrency value: %s is not valid. An integer "
+               ">= 0 must be used.\n" % concurrency)
+        stdout.write(msg)
+        return 2
     if combine:
         latest_id = repo.latest_id()
         combine_id = six.text_type(latest_id)


mtreinish:
This is what I was talking about in #202 it's independent of whatever we decide to do there for the new feature.

mtreinish:
Hmm, that's an odd failure I've not seen before: https://ci.appveyor.com/project/mtreinish/stestr/builds/21100814/job/kf7wbkoc7ug23xoo#L164 I'm going to restart that job. But it might be something we need to watch in the future.

mtreinish:
@masayukig do you have any thoughts on this?

stestr/commands/run.py
```diff
@@ -308,6 +308,11 @@ def run_command(config='.stestr.conf', repo_type='file',
             exit(1)
         repo = util.get_repo_initialise(repo_type, repo_url)
     combine_id = None
+    if concurrency and concurrency < 0:
+        msg = ("The provided concurrency value: %s is not valid. An integer "
+               ">= 0 must be used.\n")
```
masayukig: heh, 
 `">= 0 must be used.\n" % concurrency)` 
?

stestr/commands/run.py
```diff
@@ -308,6 +308,11 @@ def run_command(config='.stestr.conf', repo_type='file',
...
+        stdout.write(msg)
+        return 2
```
masayukig: And, if we run stestr with `--slowest` option, slowest test info are showed in this case. So, should we check this return value there?
https://github.com/mtreinish/stestr/pull/208/files#diff-e77977ac6a427e77fc880ed584215c86R197

stestr/commands/run.py
mtreinish: Heh, oops I'll fix that now

stestr/commands/run.py
mtreinish: Hmm, this is an interesting case. Since the return code is non-zero for expected error conditions (like a test failure) where we probably still want to print the slowest tests (when `--slowest` is set). 

I agree that we don't want to print the slowest in the case like this with an invalid input and fails before we run any tests, but I'm not sure of a way to perfectly detect that. Were you thinking check for the unique return code `2` here? I don't think all of these edge conditions will have unique return codes to detect.

masayukig: well, yeah. I see. So maybe, it's better to check it in the earlier, like after store the concurrency value?
https://github.com/mtreinish/stestr/pull/208/files#diff-e77977ac6a427e77fc880ed584215c86R172

mtreinish: Ah, yeah that's a good call I'll copy the check there too. We need to keep this inside the method too for users using the python API instead of the CLI. But, we'll never reach the inner call via the cli if the input is bad so it'll avoid the issue you're outlining here.

PR #214, opened 2019-01-04
Update docs for exit codes with --subunit flags

### LINKED ISSUE #210: stestr run testname --subunit always return 0 exist status if tests fails
### Issue description
stestr is the backend of the tempest run. when invoking tempest run -r <test regex> --subunit and the test_regex fails but the exit status is always returning 0. In tempest run, we donot use --until-failure flag in tempest run cli. 

I am seeing similar issue filed [https://github.com/mtreinish/stestr/issues/68] but it was with --subunit --until-failure, when both flag used together. It is coming with the latest master code when only --subunit flag is used.

stestr run --subunit should return 0 if tests passes and 1 if tests fails.

I am adding some tetsts in pull request to reproduce the issue: https://github.com/mtreinish/stestr/pull/209

The related issue is seen when we merged this review https://review.openstack.org/621584
and then we reverted it back https://review.openstack.org/625070

#### Specifications like the version of the project, operating system, or hardware
It is coming with master itself.

mtreinish:
The --subunit flag changes the exit code behavior in a slightly
counter intuitive manner. Basically instead of exiting non-zero when
there are failure results it only returns non-zero if subunit generation
fails. This is expected behavior, but was not clearly documented which
can cause confusion. It was only mentioned in the description for the
'last' command, and not all the commands where --subunit was available.
This commit attempts to address that by adding clear documentation to
the description string for each command. This means it will be printed
in the summary for '--help' on each command as well as ending up in the
project man page for each command.

Fixes #210

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -35,6 +35,16 @@ from stestr import user_config
 
 
 class Run(command.Command):
+    """Run the tests for a project and store them into the repository.
+
+    Without --subunit, the process exit code will be non-zero if the test
+    run was not successful. However, with --subunit, the process exit code
+    is non-zero only if the subunit stream could not be generated
+    successfully. The test results and run status are included in the
+    subunit stream, so the stream should be used to determining the result
+    of the run instead of the exit code when using the --subunit flag.
+    """
+
     def get_parser(self, prog_name):
         parser = super(Run, self).get_parser(prog_name)
         parser.add_argument("filters", nargs="*", default=None,
@@ -132,19 +142,6 @@ class Run(command.Command):
                             'attachment contents on a successful test '
                             'execution')
         return parser
-
-    def get_description(self):
-        help_str = """Run the tests for a project and store them into the
-        repository.
-
-        Without --subunit, the process exit code will be non-zero if the test
-        run was not successful. However, with --subunit, the process exit code
-        is non-zero only if the subunit stream could not be generated
-        successfully. The test results and run status are included in the
-        subunit stream, so the stream should be used to determining the result
-        of the run instead of the exit code when using the --subunit flag.
-        """
-        return help_str
 
     def take_action(self, parsed_args):
         user_conf = user_config.get_user_config(self.app_args.user_config)
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -134,8 +134,16 @@ class Run(command.Command):
         return parser
 
     def get_description(self):
-        help_str = (
-            "Run the tests for a project and store them into the repository.")
+        help_str = """Run the tests for a project and store them into the
+        repository.
+
+        Without --subunit, the process exit code will be non-zero if the test
+        run was not successful. However, with --subunit, the process exit code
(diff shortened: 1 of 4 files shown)


stestr/commands/run.py
```diff
@@ -134,8 +134,16 @@ def get_parser(self, prog_name):
         return parser
 
     def get_description(self):
-        help_str = (
-            "Run the tests for a project and store them into the repository.")
+        help_str = """Run the tests for a project and store them into the
+        repository.
```
masayukig: This change makes the `stestr help` output a bit odd.. (`list`'s description is also weird originally, though)
```
Commands:
  complete       print bash completion command (cliff)
  failing        Show the current failures known by the repository.
  help           print detailed help for another command (cliff)
  init           Create a new repository.
  last           Show the last run loaded into a repository.
  list           List the tests for a project. You can use a filter just like with the run command to see exactly what tests match
  load           Load a subunit stream into a repository.
  run            Run the tests for a project and store them into the
  slowest        Show the slowest tests from the last test run.
```

So, how about moving this to the class' docstring like this?[0]

```
class Run(command.Command):
    """Run the tests for a project and store them into the repository.

    Without --subunit, the process exit code will be non-zero if the test
    run was not successful. However, with --subunit, the process exit code
    is non-zero only if the subunit stream could not be generated
    successfully. The test results and run status are included in the
    subunit stream, so the stream should be used to determining the result
    of the run instead of the exit code when using the --subunit flag.
    """
```

> get_description()
> 
>     Return the command description.
> 
>     The default is to use the first line of the class’ docstring as the description. Set the _description class attribute to a one-line description of a command to use a different value. This is useful for enabling translations, for example, with _description set to a string wrapped with a gettext translation marker.
> 

[0] https://docs.openstack.org/cliff/latest/reference/index.html#cliff.command.Command.get_description

mtreinish: Oops, I missed your comment here, let me try moving everything to a docstring locally and test that out. If it looks good I'll update it here, otherwise I think we should do it in a follow up (and I'll comment here).

mtreinish: Yep, the docstrings are the right way to handle this. I've updated things accordingly in https://github.com/mtreinish/stestr/pull/214/commits/fef78ac369298b602fd18cef6c68550fefd93a41

mtreinish:
@masayukig when you get a chance can you fix the list command help string? Since you found the issue you should get the credit for fixing it too

masayukig:
> @masayukig when you get a chance can you fix the list command help string in a separate PR? Since you found the issue you should get the credit for fixing it too

OK, sure. So, I think init.py and slowest.py are also need to be fixed for consistency. May I do that in the PR?

mtreinish:
> > @masayukig when you get a chance can you fix the list command help string in a separate PR? Since you found the issue you should get the credit for fixing it too
> 
> OK, sure. So, I think init.py and slowest.py are also need to be fixed for consistency. May I do that in the PR?

Heh, yeah go for it.

PR #216, opened 2019-01-22
Run commands through the same python used for stestr

### LINKED ISSUE #215: The python executable used by stestr does not consider the interpreter used by stestr itself
When stestr executes additional python code, it sets the python interpreter as follow:

`python = 'python' if sys.platform == 'win32' else '${PYTHON:-python}'`

But this leads to a failure, for example, when
- stestr is executed through python3
- the alias for `python` is `python2` (as in most *nix systems nowadays)
- python2 is not available, or available but without many libraries installed (namely not subunit)
- PYTHON is not defined

then stestr execution fails with:

```
/usr/bin/python: No module named subunit        
                       
=========================                                    
Failures during discovery    
=========================
```

It would make more sense to invoke additional commands using the same interpreter used by stestr itself by default. Users should still be able to override the value by setting PYTHON as it happens nowadays.

ghost:
Tune a bit the default python interpreter used to run additional
commands: on non win32 platforms, re-use the same executable
which is used by stestr itself, unless overridden by the PYTHON
variable.

Without this fix, if (for example) stestr is executed using
python3, the execution of other commands may fail because
python is used, which usually means python2, and some dependencies
may be missing (for example subunit).

Fixes #215

### DIFF
--- a/stestr/config_file.py
+++ b/stestr/config_file.py
@@ -103,7 +103,8 @@ class TestrConf(object):
         elif not top_dir:
             top_dir = './'
 
-        python = 'python' if sys.platform == 'win32' else '${PYTHON:-python}'
+        python = sys.executable if sys.platform == 'win32' else \
+            '${PYTHON:-%s}' % (sys.executable)
         command = "%s -m subunit.run discover -t" \
                   " %s %s $LISTOPT $IDOPTION" % (python, top_dir, test_path)
         listopt = "--list"
--- a/stestr/tests/test_config_file.py
+++ b/stestr/tests/test_config_file.py
@@ -33,6 +33,7 @@ class TestTestrConf(base.TestCase):
                                expected_python='python',
                                expected_group_callback=mock.ANY):
         mock_sys.platform = platform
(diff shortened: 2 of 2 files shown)


ltoscano-rh:
I thought about sys.executable being None. But if that happens, there is probably something weird(er) going on, at least from what I read around.

mtreinish:
Yeah, I've never actually seen it return None before. The only case I've seen where sys.executable doing the unexpected thing is in the case of uwsgi, which returns the path the uwsgi binary instead of a python binary. This is because uwsgi links against python to make its own executable which then adds its own uwsgi library apis. But since stestr is pretty unlikely to get invoked by uwsgi I didn't think that failure mode was worth mentioning. But at least just gracefully handling none was something easy enough to add and have a test case for, even if it's unlikely and means something is really wrong. But this is a low priority follow-up thing it's not like I'm going to block a release on it.

ltoscano-rh:
How do you think it would be better to handle a None value? Fallback to 'python' (which may fail later, or not) or fail directly?

PR #217, opened 2019-01-23
Gracefully handle the case when sys.executable is None

ltoscano-rh:
It may happen, even if it is extremely unlikely.
Just return 'python': it may work, or just fail later - but it would
have failed anyway.

This is a follow-up of the main fix for #215.

### DIFF
--- a/stestr/config_file.py
+++ b/stestr/config_file.py
@@ -10,6 +10,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import os
 import re
 import sys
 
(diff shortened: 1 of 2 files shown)


mtreinish:
Yeah, it's an unlikely case but I still think since it's a documented possibility we should handle it instead of it just possibly returning None and raising a weird error from trying to use that as the binary name. That being said @ltoscano-rh I actually think after reviewing the code, just handling this edge condition gracefully and returning a descriptive error message saying the binary couldn't be found would be better than trying to guess what the correct binary name is.

ltoscano-rh:
But you can still force ${PYTHON} to override that value; should I check if ${PYTHON} is set too when sys.executable is empty?

mtreinish:
@ltoscano-rh Ah, yeah that's a good point, I forgot about that. Yeah I guess we should check if the env variable is set. If it is set the we don't return an error and just run it with `${PYTHON}`, if it's not set and sys.executable is empty we should raise an error. I think that behavior will be the most explicit behavior and will have the least surprises.

Issue #219, opened 2019-01-29
running test via file name doesn't work, and crashes with -n


aspiers:
### Steps to reproduce the problem

- Clone openstack/nova
- Try to run a single test file:

      tox -e py27 -- -n nova/tests/unit/virt/libvirt/fakelibvirt.py

### Expected behavior

I would expect that the tests in that file are run, as per https://stestr.readthedocs.io/en/latest/MANUAL.html#running-tests which claims:

> you can also give it a file path and stestr will convert that to the proper python path under the covers. (assuming your project don’t manually mess with import paths)

or if for some reason the file could not be located or its tests not run (e.g. if the project *does* mess with import paths - I have no idea if that is true for nova), then I expect `stestr` would give a helpful warning or even error explaining what went wrong.

### Actual behavior

`stestr` either crashes with a hopelessly cryptic error (which doesn't even look like a crash but more like a user-facing error, but I'll submit a separate issue about that):

```
$ PATH=.tox/py27/bin:$PATH stestr run -n nova/tests/unit/virt/libvirt/fakelibvirt.py                                                                
min() arg is an empty sequence
```

or, if discovery is not suppressed, silently runs zero tests with no warning that something went wrong (the `psycopg2` warning is obviously unrelated):

```
$ PATH=.tox/py27/bin:$PATH stestr run nova/tests/unit/virt/libvirt/fakelibvirt.py                                                                   
/home/adam/SUSE/cloud/OpenStack/git/nova/.tox/py27/lib/python2.7/site-packages/psycopg2/__init__.py:144: UserWarning: The psycopg2 wheel package will be renamed from release 2.8; in order to keep installing from binary please use "pip install psycopg2-binary" instead. For details see: <http://initd.org/psycopg/docs/install.html#binary-install-from-pypi>.
  """)
```

### System information

**stestr version (`stestr --version`):**

2.2.0

**Python release (`python --version`):**

2.7.14

**pip packages (`pip freeze`):**

alembic==1.0.7
amqp==2.4.0
appdirs==1.4.3
asn1crypto==0.24.0
atomicwrites==1.2.1
attrs==18.2.0
automaton==1.15.0
Babel==2.6.0
bandit==1.5.1
bcrypt==3.1.6
cachetools==3.0.0
castellan==1.1.0
certifi==2018.11.29
cffi==1.11.5
chardet==3.0.4
cliff==2.14.0
cmd2==0.8.9
colorama==0.4.1
contextlib2==0.5.5
[...]

PR #221, opened 2019-01-30
Turn on debug mode by default

### LINKED ISSUE #220: shouldn't showing stack traces (a la --debug) be the default behaviour?
## Issue description

As mentioned in #219, it is possible to cause `stestr` to crash with an error `min() arg is an empty sequence`, but with no stacktrace.  Adding `--debug` gives the stacktrace, but this begs the question - why are stack traces being hidden by default?  It gives the false impression that a crash such as this one was the fault of the user, but as a result the error message will focus on something which went wrong internally, and as a result will generally appear cryptic to the user.

### Steps to reproduce the problem

Exploit a bug such as the one mentioned in #219 in order to cause `stestr` to crash.

### Expected behavior

The stacktrace is shown, making it obvious that this was a crash not a usage error caused by the end user.

### Actual behavior

Only the error message is shown, with no indication the user hit a bug in `stestr`.

### Specifications like the version of the project, operating system, or hardware

Same as in #219.

mtreinish:
By default cliff suppresses stack traces on unhandled exceptions in order
to keep the output cleaner for users. But in practice this has caused more
confusion for stestr users since often the error messages aren't really
descriptive enough to indicate what's going wrong (or that it's even a
bug in stestr). This is often because exception messages are written
assuming the context of the stack trace for debugging, not to be user
facing messages. Cliff offers the --debug flag to enable printing the
full stack traces. This commit turns that flag on by default when stestr
initializes the cliff App object to ensure that it always prints the
full stack trace on an unhandled exception.

Fixes #220

### DIFF
--- a/stestr/cli.py
+++ b/stestr/cli.py
@@ -30,6 +30,7 @@ class StestrCLI(app.App):
             )
 
     def initialize_app(self, argv):
+        self.options.debug = True
         self.LOG.debug('initialize_app')
 
     def prepare_to_run_command(self, cmd):


masayukig:
LGTM. This can provide much more useful information to users. But I also think we should care #219 case. Maybe, another patch is needed?

When I saw this PR at the first time, I felt this is not good behavior, we should suppress stack traces to users as possible. But after I read #219 and #220 , it's OK as a kind of workaround.

So, I think an ideal behavior is that we catch all errors (in cliff or stestr), and then, print more useful message like @aspiers said https://github.com/mtreinish/stestr/issues/220#issuecomment-459015656 .

Issue #224, opened 2019-02-26
FR: history explorer for looking at old runs


aspiers:
Forgive me if I'm missing something here, but it seems that stestr goes to a fair bit of trouble to store old test runs in the file- or sqlite-backed database, but doesn't actually facilitate accessing that history in any way except `stestr last` for looking at the most recent run.  Is that right?  If so it would be really nice to have a new subcommand (e.g. `stestr history`) for exploring old runs.

## Example user stories

- As a developer who is trying to track down a heisenbug, I would like to be able to type something like `stestr history --failing=my.test.which.triggers.awkward.heisenbug` in order to obtain a list of all historical test runs where the heisenbug was triggered.

- As a developer who has made a mess and broken his local git branch, I would like to be able to find the exact time of the last run where all the tests passed, via something like `stestr history --passing --limit 1`, in order to use the git reflog to restore the code to that point.

- As a developer who is trying to pin down a strange issue stemming from the interaction of tests A and B, I would like to be able to simultaneously view in different terminals the test outputs from:
    - a passing run of just test A
    - a passing run of just test B
    - a single --serial run where test A passes and then test B fails

mtreinish:
Yeah, the history features in stestr at this point is just what testr gave us, which was never much. My intent when I forked from testr was to expand it and leverage all the features that subunit2sql provided for doing this (which is why the experimental subunit2sql backend is there). I just got side tracked and never firmed up the subunit2sql support so I never made any further progress on this. You probably could leverage this a bit today by using `subunit2sql-graph` (or even openstack-health) and the experimental sql repository type. But, that's far from where I'd like to see this.

That being said I'm definitely in favor of doing something like this, and I think the 3 use cases you've outlined are a good starting point to start discussing what this should look like and how we go about writing it. I also think we should decouple it's implementation from the subunit2sql backend (because there is still work to do on the subunit2sql side around generating an output subunit stream from the DB with attachments) which might double the future work, but at least will enable us to move forward on this.

mtreinish:
As an aside one thing you bring up is in the second example is that right now we don't track sha1 or anything from git at all. It would be cool as a simple starting point to add initial support for tracking git sha1s at `stestr run`. The only complexity comes from what about uncommitted changes? In my typical dev workflow I run tests before making a commit, if we do track the current git ref it won't really take that kind of workflow into account.

aspiers:
Cool, thanks.  Yes it should definitely be independent of the backend if possible.


[...]

PR #225, opened 2019-02-27
Add support to repositories for storing run metadata

mtreinish:
This commit adds repository APIs around storing and accessing run
metadata strings. The concept being that you can store a metadata string
to identify a run by some other identifier then the run id. The example
use case in mind for this feature is using it to store a VCS commit hash
(like a git sha1) or identifier. The APIs in this commit allow storing
metadata at run creation time, accessing the metadata for a run, and
finding run_ids by a metadata value. This is implemented for both the
file and sql repository types. The memory repository type does not have
this implemented, as it is more limited.

Related to: #224

### DIFF
--- a/stestr/repository/abstract.py
+++ b/stestr/repository/abstract.py
@@ -67,7 +67,7 @@ class AbstractRepository(object):
         """
         raise NotImplementedError(self.get_failing)
 
-    def get_inserter(self, partial=False, run_id=None):
+    def get_inserter(self, partial=False, run_id=None, metadata=None):
         """Get an inserter that will insert a test run into the repository.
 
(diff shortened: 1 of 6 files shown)


mtreinish:
This is still a work in progress because I need to add test cases. Once this feature is tested it should be good to go.

aspiers:
Coooool :-)

mtreinish:
@masayukig I just added skips for those 2 failing tests on windows. I don't know why tempfile fails there but not in other places on appveyor. But it seems like just an env issue and nothing to do with the code.

masayukig:
yeah, I don't know why either. If we can find the cause of failures, we'll be able to reenable them in the future.

PR #229, opened 2019-03-01
Update package metadata in setup.cfg

mtreinish:
After reviewing the documentation for pbr [1] and looking at pbr's own
setup.cfg[2] file there were several things to update before we push
another release. This commit goes through and updates the metadata to
match the expected format.

[1] https://docs.openstack.org/pbr/latest/user/using.html#setup-cfg
[2] https://github.com/openstack-dev/pbr/blob/master/setup.cfg

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ classifier =
     Intended Audience :: System Administrators
     Intended Audience :: Developers
     License :: OSI Approved :: Apache Software License
-    Operating System :: POSIX :: Linux
+    Operating System :: OS Independent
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -21,10 +21,11 @@ classifier =
     Programming Language :: Python :: 3.7
     Topic :: Software Development :: Testing
     Topic :: Software Development :: Quality Assurance
-project-urls =
+project_urls =
```
masayukig: I'm not sure these inconsistency - '-' vs '_' ... :(

Issue #231, opened 2019-03-04
unorderable types: str() < int()


tbreeds:
### Issue description

The 2.3.0 stestr release fails with "unorderable types: str() < int()" [1]

We believe this is due to: https://github.com/mtreinish/stestr/pull/208

### Additional information

[1] http://logs.openstack.org/79/640579/2/check/neutron-grenade-multinode/b56b183/job-output.txt.gz#_2019-03-03_06_00_19_638638

mtreinish:
I just pushed the the 2.3.1 release which includes this fix. Thanks for both identifying it and proposing the fix Tony. Sorry it took so long to address it.

stmcginnis:
"I just pushed the the 2.3.1 release which includes this fix." Did that not actually make it? According to comments in https://github.com/mtreinish/stestr/issues/235, it doesn't appear so.

mtreinish:
It did, but then ostestr was broken in another (similar) way because it was passing None (which is also not an int like the documented api) for it's concurrency value by default. The commit to translate None -> 0, #233 , was not included in the 2.3.1 release because it wasn't found until after the release.

stmcginnis:
Got it, thanks @mtreinish

PR #232, opened 2019-03-04
Ensure concurrency is always an int

tbreeds:
Closes: #231

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -34,6 +34,17 @@ from stestr.testlist import parse_list
 from stestr import user_config
 
 
+def _to_int(possible, default=0, out=sys.stderr):
+    try:
+        i = int(possible)
+    except ValueError:
+        i = default
+        msg = ('Unable to convert "%s" to an integer.  Using %d.\n' %
+               (possible, default))
+        out.write(six.text_type(msg))
+    return i
+
+
 class Run(command.Command):
     """Run the tests for a project and store them into the repository.
 
@@ -144,16 +155,6 @@ class Run(command.Command):
         return parser
 
     def take_action(self, parsed_args):
-        def to_int(possible, default=0):
-            try:
-                i = int(possible)
-            except ValueError:
-                i = default
-                msg = ('Unable to convert "%s" to an integer.  Using %d.\n' %
-                       (possible, default))
-                stdout.write(msg)
-            return i
-
         user_conf = user_config.get_user_config(self.app_args.user_config)
         filters = parsed_args.filters
         args = parsed_args
@@ -188,7 +189,7 @@ class Run(command.Command):
         verbose_level = self.app.options.verbose_level
         stdout = open(os.devnull, 'w') if verbose_level == 0 else sys.stdout
         # Make sure all (python) callers have provided an int()
-        concurrency = to_int(concurrency)
+        concurrency = _to_int(concurrency)
         if concurrency and concurrency < 0:
             msg = ("The provided concurrency value: %s is not valid. An "
                    "integer >= 0 must be used.\n" % concurrency)
@@ -330,6 +331,8 @@ def run_command(config='.stestr.conf', repo_type='file',
             exit(1)
         repo = util.get_repo_initialise(repo_type, repo_url)
     combine_id = None
+    concurrency = _to_int(concurrency)
+
     if concurrency and concurrency < 0:
         msg = ("The provided concurrency value: %s is not valid. An integer "
                ">= 0 must be used.\n" % concurrency)
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -144,6 +144,16 @@ class Run(command.Command):
         return parser
 
     def take_action(self, parsed_args):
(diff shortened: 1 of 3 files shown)


stestr/commands/run.py
```diff
@@ -177,6 +187,8 @@ def take_action(self, parsed_args):
             suppress_attachments = args.suppress_attachments
         verbose_level = self.app.options.verbose_level
         stdout = open(os.devnull, 'w') if verbose_level == 0 else sys.stdout
+        # Make sure all (python) callers have provided an int()
+        concurrency = to_int(concurrency)
```
masayukig: I think we don't need this here but in `run_command()` after
https://github.com/mtreinish/stestr/pull/232/files#diff-e77977ac6a427e77fc880ed584215c86R333

Because it must be `int` value with this.
https://github.com/mtreinish/stestr/pull/232/files#diff-e77977ac6a427e77fc880ed584215c86R62

stestr/commands/run.py
```diff
@@ -144,6 +144,16 @@ def get_parser(self, prog_name):
         return parser
 
     def take_action(self, parsed_args):
+        def to_int(possible, default=0):
+            try:
+                i = int(possible)
+            except ValueError:
+                i = default
+                msg = ('Unable to convert "%s" to an integer.  Using %d.\n' %
+                       (possible, defult))
```
masayukig: s/defult/default

stestr/commands/run.py
tbreeds: It needs to be here or in the case args.concurrency = '1' we'll set concurrency to '1' and then we'll compare '1' < 0 on line 192 and then python will complain that we're comparing a str() with an int()

masayukig: > It needs to be here or in the case args.concurrency = '1' we'll set concurrency to '1' and then we'll compare '1' < 0 on line 192 and then python will complain that we're comparing a str() with an int()

Yeah, but this should be there(on line 333) at least. Otherwise, we can't handle it like this.

>     File "/home/masayuki/git/stestr/stestr/commands/run.py", line 333, in run_command
>     if concurrency and concurrency < 0:

We actually expose `run_command()` function and tempest use it directly.

```
$ tempest --debug run  --regex '\[.*\bsmoke\b.*\]' --concurrency=4
'<' not supported between instances of 'str' and 'int'
  File "/home/masayuki/git/openstack/tempest/.tox/py37/lib/python3.7/site-packages/cliff/app.py", line 401, in run_subcommand
    result = cmd.run(parsed_args)
  File "/home/masayuki/git/openstack/tempest/.tox/py37/lib/python3.7/site-packages/cliff/command.py", line 184, in run
    return_code = self.take_action(parsed_args) or 0
  File "/home/masayuki/git/openstack/tempest/tempest/cmd/run.py", line 180, in take_action
    load_list=parsed_args.load_list, combine=parsed_args.combine)
  File "/home/masayuki/git/stestr/stestr/commands/run.py", line 333, in run_command
    if concurrency and concurrency < 0:
TypeError: '<' not supported between instances of 'str' and 'int'
  File "/home/masayuki/git/openstack/tempest/.tox/py37/bin/tempest", line 10, in <module>
    sys.exit(main())
  File "/home/masayuki/git/openstack/tempest/tempest/cmd/main.py", line 48, in main
    return the_app.run(argv)
  File "/home/masayuki/git/openstack/tempest/.tox/py37/lib/python3.7/site-packages/cliff/app.py", line 281, in run
    result = self.run_subcommand(remainder)
  File "/home/masayuki/git/openstack/tempest/.tox/py37/lib/python3.7/site-packages/cliff/app.py", line 401, in run_subcommand
    result = cmd.run(parsed_args)
  File "/home/masayuki/git/openstack/tempest/.tox/py37/lib/python3.7/site-packages/cliff/command.py", line 184, in run
    return_code = self.take_action(parsed_args) or 0
  File "/home/masayuki/git/openstack/tempest/tempest/cmd/run.py", line 180, in take_action
    load_list=parsed_args.load_list, combine=parsed_args.combine)
  File "/home/masayuki/git/stestr/stestr/commands/run.py", line 333, in run_command
    if concurrency and concurrency < 0:
TypeError: '<' not supported between instances of 'str' and 'int'
```

tbreeds: Ahh okay I see.  I need it in both places.  I'll re-spin tomorrow if not beaten to it ;P

masayukig: > Ahh okay I see. I need it in both places. I'll re-spin tomorrow if not beaten to it ;P

heh. But well, I think we don't need it here because the value type must be int.

```
        parser.add_argument("--concurrency", action="store", default=None,
                            type=int,
                            help="How many processes to use. The default (0) "
                            "autodetects your CPU count.")
```

mtreinish: I've updated the PR to do the checking in both places for the CLI and the Python API so we can get 2.3.1 out the door today.

@masayukig I think you're right Cliff/argparse will do the type enforcement for us, but I left it in place because I think the function will give us a good place to handle #202 when we decide on what the grammar should be for that.

PR #233, opened 2019-03-06
Take care of 'concurrency is None' case

masayukig:
This commit fixes `concurrency is None` case. The original code was
taking care of ValueError case such as 'foo', but if we pass `None` to
the `_to_int()`, it raised a TypeError. So, this commit converts it to a
ValueError instead.

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -37,7 +37,7 @@ from stestr import user_config
 def _to_int(possible, default=0, out=sys.stderr):
     try:
         i = int(possible)
-    except ValueError:
+    except (ValueError, TypeError):
         i = default
         msg = ('Unable to convert "%s" to an integer.  Using %d.\n' %
                (possible, default))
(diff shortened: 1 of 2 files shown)


masayukig:
ah, I should mention #231 in the commit message...

stestr/commands/run.py
```diff
@@ -36,6 +36,8 @@
 
 def _to_int(possible, default=0, out=sys.stderr):
     try:
+        if possible is None:
+            raise ValueError(possible)
         i = int(possible)
     except ValueError:
```
mtreinish: Couldn't we just add ValueError to this exception with something like:

```suggestion
    except (ValueError, ValueError):
```

instead of converting the exception type?

mtreinish:
I'm also debating whether this warrants a 2.3.2 release. I don't think it does, maybe we should just fix tempest to unblock global-requirements

stestr/commands/run.py
masayukig: Ah, yeah, that must be simpler :) I'll update it soon.

masayukig:
> I'm also debating whether this warrants a 2.3.2 release. I don't think it does, maybe we should just fix tempest to unblock global-requirements

I pushed a patch to set a default value for `tempest run` command. https://review.openstack.org/641349

PR #239, opened 2019-03-27
Update copyright year

masayukig:
This commit updates the copyright year in the document. It's better to
keep it up-to-date because it prevents people think stestr is stale or
dead with the documents.


mtreinish:
Sure I guess this is fine. Do people really read that much into the copyright date there though?

Issue #223, opened 2019-02-19
Add flag to always print all text attachments


aspiers:
**This issue description was inaccurate - please read the whole issue**

## Issue description

When debugging tests it's sometimes desirable to print all text attachments from tests even during success. The default behavior right now is to print STDOUT and STDERR for successful tests and all text attachments for failed tests. There is also a `--suppress-attachments` flag that when set will not print any output for successful tests. It would be good to have an option/flag added that is the opposite of ``suppress-attachments` which when set will print all attachments, not just STDOUT and STDERR.

### Previous description:

https://stestr.readthedocs.io/en/latest/MANUAL.html#adjusting-test-run-output says

>In the default subunit-trace output any captured output to stdout and stderr is printed after test execution, for both successful and failed tests.

but the behaviour I see is that it is only shown for failed tests.

#### Steps to reproduce the problem

e.g. with nova:

```
$ source .tox/functional/bin/activate
$ stestr --test-path=./nova/tests/functional run nova.tests.functional.test_servers
```

and then I see a load of passing tests with no captured output, e.g.

```
...
{7} nova.tests.functional.test_servers.ServerMovingTestsWithNestedResourceRequests.test_resize_reschedule_uses_host_lists_1_fail [6.512132s] ... ok
{0} nova.tests.functional.test_servers.ServersTest.test_create_server_with_error [3.748631s] ... ok
{4} nova.tests.functional.test_servers.ServerMovingTestsWithNestedResourceRequests.test_resize_instance_fails_allocation_cleanup [11.451513s] ... ok
{6} nova.tests.functional.test_servers.ServerMovingTestsWithNestedResourceRequests.test_live_migrate_rollback_cleans_dest_node_allocations [5.875147s] ... ok
{5} nova.tests.functional.test_servers.ServerMovingTestsWithNestedResourceRequests.test_evacuate_host_specified_but_not_forced [8.125972s] ... ok
{1} nova.tests.functional.test_servers.ServerRebuildTestCase.test_rebuild_with_new_image [4.712644s] ... ok
{3} nova.tests.functional.test_servers.ServerMovingTestsWithNestedResourceRequests.test_evacuate [7.441238s] ... ok
{2} nova.tests.functional.test_servers.ServerSoftDeleteTestsWithNestedResourceRequest.test_soft_delete_then_restore [7.302989s] ... ok
{0} nova.tests.functional.test_servers.ServersTest.test_deferred_delete_restore_overquota [5.372631s] ... ok
[...]

PR #240, opened 2019-04-05
Print attachments when a test finishes successfully

masayukig:
This commit fixes to print attachments which include captured stdout and
stderr when a test finishes successfully.

Closes: #223


mtreinish:
Sorry, I missed this before. I'm not sure this is what we want to do. I think we need to discuss what the actual bug is in #223 before we merge this.

masayukig:
I'll close this PR because I don't think this is desirable anymore. Actually, I was also think that adding `--show-attachment` option, but it might not be needed so far at least.

PR #241, opened 2019-05-14
Discover python executable when discover is not used

amoralej:
Currently it's hardcoded to 'python' wich in some cases may not exist,
as in some distros with python3 only.

This patch uses PYTHON env variable if defined or executable used to run
stestr.


mtreinish:
Hmm, it looks like travis completed successfully but for some reason the check is still listed as in progress. I'm going to override it because that's clearly wrong: https://travis-ci.org/mtreinish/stestr/builds/532305786?utm_source=github_status&utm_medium=notification

PR #247, opened 2019-07-15
Rework logic for suppress attachments and all attachments

mtreinish:
This commit reworks the logic for handling the 2 mutually exclusive
options suppress attachments and all attachments. Previosuly supress
attachments would always take precedence over all attachments, even if
all attachments were set via the cli and all attachments were in the
user config file. This commit makes the behavior more consistent with
expectations and has it so cli arguments always take precedence. It also
changes the behavior to error if both options are set on the cli.

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -210,10 +210,12 @@ is also possible that a test has other text attachments (a common example is
 python logging) which are not printed on successful test execution, only on
 failures. If you would like to have these attachments also printed for
 successful tests you can use the ``--all-attachments`` flag to print all text
-attachments on both successful and failed tests. If both ``--all-attachments``
-and ``--suppress-attachments`` are set then the ``--suppress--attachments``
-flag will take priority and no attachments will be printed for successful
-tests.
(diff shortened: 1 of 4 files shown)


doc/source/MANUAL.rst
```diff
@@ -210,10 +210,12 @@ is also possible that a test has other text attachments (a common example is
...
-attachments on both successful and failed tests. If both ``--all-attachments``
-and ``--suppress-attachments`` are set then the ``--suppress--attachments``
-flag will take priority and no attachments will be printed for successful
-tests.
+attachments on both successful and failed tests. Both ``--all-attachments``
+and ``--suppress-attachments`` can not be set at the same time. If both are
+set in the user config file then the ``suppress-attachments`` flag will take
+priority and no attachments will be printed for successful tests. If either
+``--suppress-attachments`` or ``--all-attachments`` is set via the CLI it
+will take precedence over matching options set in the user config file.
```
masayukig: I feel the difference between CLI options and user configs is a bit confusing. But I think it's reasonable. Because users can recognize the CLI options easily but not user configs.

PR #248, opened 2019-07-15
Add readme note about python 2 eol in readme

mtreinish:
Now that we're getting ready to drop python 2 support for an stestr
3.0.0 release in early 2020 this commit adds a release note to make the
plans clear to end users.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -20,6 +20,10 @@ stestr
 * You can see the full rendered docs at: http://stestr.readthedocs.io/en/latest/
 * The code of the project is on Github: https://github.com/mtreinish/stestr
 
+.. note:: stestr v2.x.x release series will be the last series that supports
+    Python 2. Support for Python 2.7 will be dropped in stestr release 3.0.0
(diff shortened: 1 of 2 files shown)


mtreinish:
@masayukig I was also debating adding a deprecation warning if python2 was used. But I'm wondering if that's too noisy for users. Do you have any thoughts?

masayukig:
Yeah, I think adding a deprecation warning is good. It might be a bit too noisy but it's necessary to advertise.

masayukig:
This works in my environment at least.
```
$ PYTHONWARNINGS=all stestr run
/home/masayuki/git/stestr/.tox/py27/bin/stestr:10: DeprecationWarning: Python 2.7 will reach the end of its life on January 1st, 2020. Support for using python 2.7 with stestr will be removed in the 3.0.0 release in early 2020
 :
```

mtreinish:
@masayukig Is there anything else left to do on this?

masayukig:
Nope. I think we can merge this now.

PR #250, opened 2019-07-15
Add release automation for pypi artifacts

mtreinish:
This commit adds travis jobs on tag releases to build the release
artifacts for pypi. It will generate both an sdist and a universal
wheel and then upload that to pypi. This automates the release process
so that when it's release time we only need to push a tag to github
and everything else is handled automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,15 @@ matrix:
       env: TOXENV=docs
     - python: "2.7"
       env: TOXENV=py27
+    - if: tag IS present
+      python: "3.6"
+      env:
+        - TWINE_USERNAME=stestr-release
+      install: pip install -U twine
+      script:
+        - python3 setup.py sdist bdist_wheel --universal
+        - twine upload dist/stestr*
+
 cache:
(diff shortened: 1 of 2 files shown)


mtreinish:
Putting this on hold until I can create a bot account in pypi to leverage this. When I do I'll add the password to the travis secret list and remove the `on hold` tag

mtreinish:
Ok, I've removed the on hold label now because I've created a bot account and added the password to the travis secrets list. It should be populated for the job when we push a tag in the future.

.travis.yml
```diff
@@ -19,6 +19,15 @@ matrix:
       env: TOXENV=docs
     - python: "2.7"
       env: TOXENV=py27
+    - if: tag IS present
+      python: "3.6"
```
masayukig: I was curious why not 3.7 here. But 3.7 requires special setup. So, 3.6 is OK here.

masayukig:
We can't see the code works properly until we push a tag. But I think it should work.

PR #251, opened 2019-07-16
Set up CI with Azure Pipelines

mtreinish:


azure-pipelines.yml
```diff
@@ -0,0 +1,34 @@
...
+    Python37:
+      python.version: '3.7'
+
+steps:
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '$(python.version)'
+  displayName: 'Use Python $(python.version)'
+
+- script: |
```
masayukig: We need to run when the python.version == '2.7'.
```
choco install vcpython27 --yes
```
According to https://github.com/microsoft/azure-pipelines-image-generation/issues/793

masayukig:
Maybe, the next step is moving travis jobs to Azure pipelines?

mtreinish:
I think we should just start by doing windows testing on this. Once we've used it for a little while we can look at fully replacing travis too.

masayukig:
Yeah, I agree. Let's see the stability, performance, etc.
Do you have anything else to do for this pull request?

PR #252, opened 2019-07-17
Unskip get/find metadata tests

masayukig:
This commit makes the tests unskip since we moved windows tests to the
Azure pipeline environment. The root cause of failure was unknown on the
appveyor environment. So, these would work in the new environment
hopefully.


masayukig:
hmm, this doesn't work by the same error.

https://dev.azure.com/stestr/stestr/_build/results?buildId=7&view=logs&jobId=76ed1102-9071-5354-720f-6f09bbc9f0c0&taskId=be5b5b08-9bc6-5c95-c335-5e65f8440ac3&lineStart=1566&lineEnd=1567&colStart=1&colEnd=1

Issue #249, opened 2019-07-15
Switch windows testing to azure pipelines instead of appveyor


mtreinish:
### Issue description

For work I recently switched projects away from using appveyor for windows testing to azure pipelines (https://github.com/marketplace/azure-pipelines ). It's a much more capable platform and the best part is the capacity available for open source projects is 10x that of appveyor. To make testing faster it would be good to look at switching our windows testing to use azure pipelines instead.

It might also be worth considering replacing our travis usage with azure pipelines too since it supports windows, osx, and linux and the 10 concurrent jobs exceeds the 3 we get from travis too.

masayukig:
This was completed by #251 ? (partially?) We don't replace the all of the travis usage yet but windows testing was already switched.

masayukig:
> It might also be worth considering replacing our travis usage with azure pipelines too since it supports windows, osx, and linux and the 10 concurrent jobs exceeds the 3 we get from travis too.

Oh, it supports OSX. That means we can have testing for the environment finally :)

masayukig:
Can we close this issue?

mtreinish:
Yeah, this is done for now. If we decide to migrate other things over to azure in the future we can do that or open a new issue for it.

PR #253, opened 2019-07-17
Add macOS py37 azure pipelines

masayukig:
This commit adds a macOS Mojave(10.14) testing job strategy to the azure
pipelines. In the document[0], it supports macOS 10.14 and 10.13 but we
don't need to test all versions of macOS and Python. So, as a first
step, this commit just add one job.

[0] https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#use-a-microsoft-hosted-agent

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,33 +6,57 @@
 trigger:
 - master
 
-pool:
-  vmImage: 'vs2017-win2016'
-strategy:
-  matrix:
-    Python27:
-      python.version: '2.7'
-    Python35:
-      python.version: '3.5'
-    Python36:
-      python.version: '3.6'
-    Python37:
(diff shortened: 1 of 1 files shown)


masayukig:
This is a partial solution of #249

masayukig:
hmm, It looks github doesn't recognize the multiple commits branch on my repo..

masayukig:
> I'm thinking we should make 2 changes here. First lets add all python versions on macos. I agree we only need 10.14 the differences between osx version are of less concern, but making sure all python versions on python work I think is important. The other change I think would should make is changing the structure here slightly, I would do:

Thanks for your comment! yeah, make sense. I changed the code as you suggested.

PR #258, opened 2019-07-24
Add unit tests for output functions

masayukig:
This commit adds unit tests for output functions to increase test
coverage and ensure their outputs.


masayukig:
> LGTM, thanks for improving the test coverage. I'm curious why you didn't use io.BytesIO instead of a tempfile so we could keep it in memory, but it's not a blocker.

ah, yeah, I should have used it. I'll push a patch for that soon.

PR #256, opened 2019-07-20
Add custom test runner and remove testtools runner dependency

mtreinish:
Right now stestr doesn't actually own it's own test runner. It leverages
the subunit.run test runner, subunit.run in turn inherets from
testtools.run, which inherets from unittest2. The usage of unittest2
causes all sorts of problems when running non-testtools or non-unittest2
test suites, because of mismatches between stdlib unittest and unittest2.
To address this problem this commit creates a new test runner for stestr
that will emit subunit which is the actual only hard requirement for the
internal test runner used. This new test runner is more or less identical
to subunit.run, but instead of inheriting from testtools it uses stdlib
unittest directly.
    
Python 2.7's stdlib unittest's test runner is much older and it was
changed significantly for python 3. So building a custom test runner
based on python 2's unittest would basically requiring maintaining a
completely separate copy of python 3's unittest runner in tree and then
building the subunit runner off of that, which is essentially what
unittest2 is anyway. Since python 2 support is deprecated anyway and
will be going away soon this commit just maintains the current status
quo of basing it off testtools for python 2.

Most of the code for the custom runner and the TestProgram class are
taken directly from subunit and testtools respectively.
    
Fixes #244 for the first stacktrace outlined there, which was caused by
the test runner (which was previously outside the scope of stestr).

PR #260, opened 2019-07-25
Use StringIO instead of tempfile

masayukig:
This commit makes unit tests in test_output.py use io.StringIO instead
of tempfile. It can reduce the code (no need seek(0)). And this commit
also fixes inconsistent code in output.py::output_tests() because it
caused an error with StringIO and it should be the same with the later
one.
The last thing is fixing the indentation in output_stream() which was
just weird.

### DIFF
--- a/stestr/output.py
+++ b/stestr/output.py
@@ -75,7 +75,7 @@ def output_tests(tests, output=sys.stdout):
 
     for test in tests:
         id_str = test.id()
-        output.write(id_str)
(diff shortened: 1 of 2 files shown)


mtreinish:
Oh interesting the azure pipelines python3.6 job failed here too. I guess that means it's not something I introduced in #256. That's good to know it's something larger that changed in CI and I can revert my attempts to fix the possible causes of the failure in that PR.

PR #261, opened 2019-07-31
Add README_ja.rst

masayukig:
This commit adds README_ja.rst which is a Japanese translation of
README.rst. This might be helpful to encourage to use stestr and to
participate development for Japanese.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -17,8 +17,12 @@ stestr
     :target: https://pypi.python.org/pypi/stestr
     :alt: Latest Version
 
+* Read this in other languages: `English`_, `日本語`_
 * You can see the full rendered docs at: http://stestr.readthedocs.io/en/latest/
 * The code of the project is on Github: https://github.com/mtreinish/stestr
+
+.. _English: https://github.com/mtreinish/stestr/blob/master/README.rst
+.. _日本語: https://github.com/mtreinish/stestr/blob/master/README_ja.rst
 
 .. note:: stestr v2.x.x release series will be the last series that supports
     Python 2. Support for Python 2.7 will be dropped in stestr release 3.0.0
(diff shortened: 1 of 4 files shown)


masayukig:
> LGTM, I'll defer to you on the actual translation. But, the only thing I think we should make sure to do here is link to the translated version off the main `README.rst` fail I would just add a line like:
> 
> ```
> * Read this in other languages: `English`_, `日本語`_
> 
> .. _English: https://github.com/mtreinish/stestr/blob/master/README.rst
> .. _日本語: https://github.com/mtreinish/stestr/blob/master/README_ja.rst
> ```
> 
> right below the links at the start of the README. That way it'll be clear where to find to the translated version.

Thanks. Yeah, that looks better. I'll push a patch :)

mtreinish:
We probably should also add the equivalent line to the ja version too (so it links back to English)

masayukig:
yeah, that makes sense.

masayukig:
So, should we have the English version link in README.rst? I think that should be the same as a user looking at.

.. _English: https://github.com/mtreinish/stestr/blob/master/README.rst

masayukig:
Yeah, I agree with that the ja README in the en docs is a bit confusing. But as a starting point to have translated copies, I think just adding a translated README is not so bad. To be more generic/sophisticated, I need to know more readthedocs.io, zanata, sphinx, pot/po files etc. It shouldn't be so hard but I'm very new to like those things yet. 

As a next step, I made the locale and put the translated file[0]. We can generate Japanese documents like this.
```
sphinx-build -c doc/source -b html -D language=ja  doc/source doc/build
```

[0] https://github.com/masayukig/stestr/tree/i18n/doc/source/locale/ja/LC_MESSAGES

PR #262, opened 2019-07-31
Add linux azure pipelines

masayukig:
This commit adds a linux (Ubuntu 16.04) testing job strategy to the
azure pipelines. This commit doesn't remove the current travis jobs
because it still works well and we should look at its stability of the
azure one. So, the next step would be removing the duplicated job
definition and implementing coveralls execution in the azure-pipelines.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -60,3 +60,28 @@ jobs:
     displayName: 'Install dependencies'
   - script: tox -e py
     displayName: 'Run Tox'
+
+- job: 'Linux_Tests'
+  pool: {vmImage: 'ubuntu-16.04'}
+  strategy:
+    matrix:
+      Python27:
+        python.version: '2.7'
+      Python35:
+        python.version: '3.5'
+      Python36:
+        python.version: '3.6'
+      Python37:
+        python.version: '3.7'
+  steps:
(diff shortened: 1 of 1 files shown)


mtreinish:
So the run time of the jobs here were roughly the same (which is inline with my experience) but where azure pipelines normally provides an advantage is on queuing time normally (we get 10x parallel jobs and the nodes launch much quicker). Do you know how much quicker azure was to finish than travis? Also, what were you thinking for a long term migration plan, we use travis for more than just unittest jobs were you thinking we'd migrate those too? Or will we keep travis around for a class of jobs too?

masayukig:
> Do you know how much quicker azure was to finish than travis?

Well, I actually don't know how much quicker. So, I wanted to check it with this patch.

>  what were you thinking for a long term migration plan, we use travis for more than just unittest jobs were you thinking we'd migrate those too? Or will we keep travis around for a class of jobs too?

Yes, we use travis for more than that. I was thinking we'd migrate those too if it has advantages. However, it looks there isn't that much. - The run time of the jobs are almost the same.

So, my conclusion is keep the travis jobs and not to add Linux tests to azure pipeline at this point. In the future, when we need to have more conbinations of python/os versions, we'll be able to revisit here.

PR #263, opened 2019-08-05
Handle NoneType timestamps in subunit results

mtreinish:
For some reason there are cases when a Nonetype object gets set as a
timestamp for a subunit result. Then when we go to analyze the total
test run in stestr load for subunit-trace output the min() or max()
calls to figure out the total test run duration it will fail because
python can't compare a Nonetype and a datetime.datetime. This commit
addresses this by ignoring cases where the a None is inserted.


mtreinish:
Merging this quickly to workaround a bug in ci for qiskit-terra. I'll come back and add unit tests in the near future.

PR #265, opened 2019-08-16
Only use stdlib based unittest runner for py >=3.5

### LINKED ISSUE #264: Issue with stestr 2.5.0 using Python 3.4
### Issue description

On one of our projects for which we use `tox` to run our tests against different versions of Python, our Python 3.4 test is consistently failing. From the following trace, we identified the issue came [from this line](https://github.com/mtreinish/stestr/commit/c50a9a70b3a723d4ccaf12b0dfc4a6a745ec2337#diff-b1dc26084457524f8efa9922f12e0bb0R65) added in https://github.com/mtreinish/stestr/commit/c50a9a70b3a723d4ccaf12b0dfc4a6a745ec2337 between 2.4.0 and 2.5.0.

Screenshot 2019-08-15 at 09 38 45

We are aware of Python 3.4 deprecation, as per [PEP-429](https://www.python.org/dev/peps/pep-0429/) but since we are shipping a library, we cannot deprecate Python 3.4 immediately.

- Expected behavior: no error running `stestr` with Python 3.4
- Actual behavior: no error running `stestr` with Python 3.4

- Run a test with Python 3.4 using `stestr` at version 2.5.0

#### Specifications like the version of the project, operating system, or hardware

- Python 3.4
- MacOS and Linux
- `stestr` 2.5.0

### System information

**stestr version (`stestr --version`):**
2.5.0

**Python release (`python --version`):**
Python 3.4

**pip packages (`pip freeze`):**
N/A

mtreinish:
The stdlib python test runner introduced in #256 was written assuming a
python stdlib from python 3.5 or newer. Since we only officially support
running on python 3.5 or newer when we added the logic to switch between
the stdlib runner and the old testtools based runner we just said
python >=3.0.0. However, despite our official support status for
python >2.7,<3.5 there are still users out there running stestr on these python
versions. The new test runner will not work for those users because
unittest in stdlib changed in python 3.5 (as an aside this is the same
issue I've been hitting in my attempts to get unittest2 removed from
testtools). To address this problem a simple fix is to just change the
if statement to be a bit more explicit and only use the new runner for
python 3.5 or newer. This way stestr is more explicit about where the
newer runner actually works and this should hopefully unbreak any users
out there still using python 3.0 to 3.4.

Fixes #264

### DIFF
--- a/stestr/subunit_runner/run.py
+++ b/stestr/subunit_runner/run.py
@@ -84,7 +84,7 @@ class SubunitTestRunner(object):
 
 def main():
     runner = SubunitTestRunner
-    if sys.version_info[0] >= 3:
+    if sys.version_info[0] >= 3 and sys.version_info[1] >= 5:
         program.TestProgram(
             module=None, argv=sys.argv,
             testRunner=partial(runner, stdout=sys.stdout))


stestr/subunit_runner/run.py
```diff
@@ -84,7 +84,7 @@ def _list(self, test):
 
 def main():
     runner = SubunitTestRunner
-    if sys.version_info[0] >= 3:
+    if sys.version_info[0] >= 3 and sys.version_info[1] >= 5:
```
masayukig: It might be better to add a comment/note to understand the magic number '5' here. But I think it's not a blocker for merging because we can check it with git log/blame.

masayukig:
I'm sorry for my delayed review. LGTM

Issue #269, opened 2019-09-05
Windows paths aren't properly handled


petrutlucian94:
### Issue description

stestr cannot handle Windows paths having trailing backslashes or white spaces. Trailing backslashes will be treated as escape characters, in which case the commands will not be properly interpreted.

Trace: http://paste.openstack.org/raw/771273/

#### Expected behavior and actual behavior

stestr should be able to run on Windows. There already are some tests aiming to ensure Windows compatibility.

#### Steps to reproduce the problem

* use "stestr list" pointing to a path that has a trailing backslash
* use a Python binary located at a path containing a white space

#### Specifications like the version of the project, operating system, or hardware

master branch (>2.5.1)

### System information

**stestr version (`stestr --version`):**

**Python release (`python --version`):**

**pip packages (`pip freeze`):**

PR #270, opened 2019-09-05
Escape Windows paths

petrutlucian94:
Windows uses backslash as path separator. Trailing backslashes
have to be escaped in order to avoid having the commands incorrectly
interpreted.

At the same time, it's quite common for Windows paths to contain
whitespaces, for which reason we'll have to use quotes.

Closes: #269

### DIFF
--- a/stestr/tests/test_config_file.py
+++ b/stestr/tests/test_config_file.py
@@ -10,12 +10,14 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+import ddt
 import mock
 
 from stestr import config_file
 from stestr.tests import base
 
 
+@ddt.ddt
 class TestTestrConf(base.TestCase):
 
     @mock.patch.object(config_file.configparser, 'ConfigParser')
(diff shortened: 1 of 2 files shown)


stestr/tests/test_config_file.py
```diff
@@ -49,7 +51,7 @@ def _check_get_run_command(self, mock_sys, mock_TestProcessorFixture,
         self.assertEqual(mock_TestProcessorFixture.return_value, fixture)
         mock_get_repo_open.assert_called_once_with('file',
                                                    None)
-        command = '%s -m stestr.subunit_runner.run discover -t "%s" "%s" ' \
+        command = '"%s" -m stestr.subunit_runner.run discover -t "%s" "%s" ' \
```
masayukig: It looks this change causes the error of `tox -e cover` because of `PYTHON=coverage run --source stestr` here.

https://travis-ci.org/mtreinish/stestr/jobs/581131380#L248
https://github.com/mtreinish/stestr/blob/master/tox.ini#L28

petrutlucian94: Makes sense. So the callable may contain white spaces either because of the path or because of extra arguments. One reasonable workaround would be to check if the path exists, if so we can keep the quotes, otherwise we can assume that extra arguments are being used and the quotes should be dropped.

petrutlucian94:
> I'm not sure about the errors on py36.

I think we're hitting this: https://bugs.python.org/issue37380. It could be that a process fails to spawn and with a little bit of luck, it may go away if we'll correctly handle the case in which the python binary path includes extra arguments.

mtreinish:
@masayukig Nah, it looks fine to me.

Issue #14, opened 2017-01-03
Add a --pdb option to stestr run


mtreinish:
Right now if you want to run with pdb stestr breaks that. The way to workaround this is to call subunit.run or testtools.run directly. We point this out in the stestr development docs:

http://stestr.readthedocs.io/en/latest/developer_guidelines.html#running-the-tests

However there is no reason we can't just fast path stestr to call the testr runner directly if we want to use pdb. We probably don't want to store results in the repository in that case.

For an example of this being done, we added this to os-testr: https://github.com/openstack/os-testr/blob/master/os_testr/ostestr.py#L197-L210 we probably could steal that logic and dump it into stestr.commands.run.

mtreinish:
So we added the - -no-discover option already which does the same thing and calls subunit.run directly. All we need to do here is add to the docs around pdb usage to tell people what to do.

mtreinish:
I actually don't know if --no-discover will actually work for using pdb. It still subprocesses out which might cause an issue with pdb. Someone will need to test it works correctly with --no-discover before we call it implemented and just document how to do it.

aspiers:
I just tested it and it works:

    $ stestr --test-path=./nova/tests/functional run -n nova.tests.functional.test_servers.TraitsTrackingTests.test_resource_provider_traits
    > /home/adam/SUSE/cloud/OpenStack/git/nova/.tox/functional/lib/python2.7/site-packages/placement/requestlog.py(44)__call__()
    -> accept = environ.get('HTTP_ACCEPT')
    (Pdb)

In contrast:

    $ stestr --test-path=./nova/tests/functional run TraitsTrackingTests
    > /home/adam/SUSE/cloud/OpenStack/git/nova/.tox/functional/lib/python2.7/site-packages/placement/requestlog.py(44)__call__()
    -> accept = environ.get('HTTP_ACCEPT')
    (Pdb)
    {0} nova.tests.functional.test_servers.TraitsTrackingTests.test_resource_provider_traits [4.816931s] ... FAILED

    Captured traceback:
    ~~~~~~~~~~~~~~~~~~~
        Traceback (most recent call last):
          File "/home/adam/SUSE/cloud/OpenStack/git/nova/.tox/functional/lib/python2.7/site-packages/mock/mock.py", line 1724, in _inner
            return f(*args, **kw)
          File "nova/tests/functional/test_servers.py", line 2047, in test_resource_provider_traits
            self._delete_trait('CUSTOM_FOO')
[...]

PR #271, opened 2019-09-08
[WIP] Add dynamic scheduling operation mode

mtreinish:
This commit adds an opt-in scheduler option for dynamic scheduling.
Instead of partitioning the test list up-front based on historical
timing data this commit lets each worker ask for the next test
dynamically.  This is built using python's multiprocess module to
launch new workers instead of shelling out to call python via
subprocess.

This hopefully will provide a better worker balance since we will keep
each worker occupied until there are no more tests to be run. Instead
of trying to pack fill each work optimially up front. Additionally this
should hopefully improve the pdb story for users who use pdb with tests.
Since instead of spawning subprocesses calling python to invoke the
subunit runner and reading the subunit stream from stdout and instead
uses multiprocessing to fork workers and uses pipes to pass the subunit
streams between workers.


mtreinish:
This still fails some tests but after switching away from `subunit.run` this seems to run reliably on python >=3.5. It also will need proper test coverage since this flow is very different from what was there before. I probably should add a dynamic job to travis and appveyor to split out running the full suite between the old way and this new approach.

stestr/commands/run.py
```diff
@@ -363,6 +366,13 @@ def run_command(config='.stestr.conf', repo_type='file',
                ">= 0 must be used.\n" % concurrency)
         stdout.write(msg)
         return 2
+    if dynamic:
+        if sys.version_info()[0] < 3 or sys.version_info[1] < 5:
```
aspiers: For obvious reasons this causes:

    TypeError: 'sys.version_info' object is not callable

```suggestion
        if sys.version_info[0] < 3 or sys.version_info[1] < 5:
```

mtreinish: Oops, good catch thanks

aspiers (commented):
Thanks for accepting my suggestion.  As per [this comment](https://github.com/mtreinish/stestr/issues/14#issuecomment-529433109) this doesn't seem to work for me yet.

thomasgoirand:
Under Python 3.13, I get:

```
  File "/usr/lib/python3/dist-packages/stestr/output.py", line 183, in __del__
    self.proc.join()
AttributeError: 'dict' object has no attribute 'join'
  File "<string>", line 1, in <module>
    from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=6, pipe_handle=103)
                                                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/multiprocessing/spawn.py", line 122, in spawn_main
    exitcode = _main(fd, parent_sentinel)
  File "/usr/lib/python3.13/multiprocessing/spawn.py", line 132, in _main
    self = reduction.pickle.load(from_parent)
  File "/usr/lib/python3.13/multiprocessing/synchronize.py", line 115, in __setstate__
    self._semlock = _multiprocessing.SemLock._rebuild(*state)
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory
```

Issue #266, opened 2019-08-22
stestr run doesn't handle an empty .stestr dir


mtreinish:
### Issue description

When you run `stestr run` and there is an empty .stestr/ directory in the run dir this raises an unhandled exception. This is because stestr run only checks for that the repository path exists not that it's a valid repository. For example:
```pytb
Traceback (most recent call last):

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/repository/file.py", line 54, in open

    stream = open(os.path.join(base, 'format'), 'rt')

FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/mtreinish/aqua/.stestr/format'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/commands/run.py", line 348, in run_command

    repo = util.get_repo_open(repo_type, repo_url)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/repository/util.py", line 47, in get_repo_open

    return repo_module.RepositoryFactory().open(repo_url)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/repository/file.py", line 57, in open

    raise repository.RepositoryNotFound(url)

stestr.repository.abstract.RepositoryNotFound: No repository found in /home/travis/build/mtreinish/aqua. Create one by running "stestr init".

During handling of the above exception, another exception occurred:

Traceback (most recent call last):

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/cliff/app.py", line 401, in run_subcommand

    result = cmd.run(parsed_args)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/cliff/command.py", line 185, in run

    return_code = self.take_action(parsed_args) or 0

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/commands/run.py", line 235, in take_action

    all_attachments=all_attachments)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/commands/run.py", line 357, in run_command

    repo = util.get_repo_initialise(repo_type, repo_url)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/stestr/repository/util.py", line 68, in get_repo_initialise

[...]

PR #273, opened 2019-09-09
Handle empty .stestr directory to initialize

masayukig:
This commit make an empty .stestr directory initialize if it exists.
Originally, we make it errors or weird errors occur when there is.
However, an empty .stestr directory can be initialized and no error
shouldn't be occurred with it.

Fixes: #266

### DIFF
--- a/stestr/commands/init.py
+++ b/stestr/commands/init.py
@@ -12,6 +12,7 @@
 
 """Initialise a new repository."""
 
+import errno
 import sys
 
 from cliff import command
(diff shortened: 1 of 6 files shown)


masayukig:
@mtreinish any thoughts?

masayukig:
> This LGTM, but I think having some tests to cover these new conditions would be useful. The logic around handling this is getting more complicated.

Thanks. I agree with you. This should have some tests because of the complication. I'll do.

masayukig:
@mtreinish : I added some unit tests. What do you think?

PR #276, opened 2019-10-07
Add --pdb flag to stestr run

### LINKED ISSUE #14: Add a --pdb option to stestr run

mtreinish:
This commit adds a new flag --pdb to stestr running a test id (in a
similar fashion to --no-discover) for when pdb is going to be invoked
from the tests. Pdb has trouble running in situations where another
process is launched (which is pretty much all of stestr, including with
--no-discover). This commit adds a --pdb flag so that you can launch
the test runner without any additional processes being launched. This
will run the tests in a process capture the subunit output to an buffer
and after the tests finish executing will pass that to the load command
so it can write it to the repository and generate subunit-trace or any
other output format (per the set flags). The tradeoff here is that the
output will not be real time from the test instead it will only be
processed until after the test executes.

One thing to note is that if you're using a fixture to capture stdout
that will sometimes take the output from pdb so you will be in the
debugger but get no output. If you plan to debug a test suite with a
fixture make sure to disable that first.

Fixes #14


mtreinish:
@aspiers  can you take a look at this too to make sure it actually improves the pdb use case for you? I tested it locally and was able to verify that pdb was invoked correctly when setting a breakpoint in the code. But I don't typically use pdb in my debugging so I'm not sure if I'm missing something else.

stestr/commands/run.py
```diff
@@ -429,6 +454,29 @@ def run_tests():
                 if result:
                     return result
 
+    if pdb:
+        ids = pdb
+        if '::' in ids:
+            ids = ids.replace('::', '.')
+        if ids.find('/') != -1:
```
masayukig: '/' can be '\' in Windows case? But this is a different topic.

mtreinish: Oh, that's a good point.  I just copied this section from `- - no-discover`'s block. Yeah I agree we should probably refractor this section in a follow up pr.

aspiers:
Unlikely to be able to do this soon, sorry.  But please test interactive usage (completion etc.) with pdbpp if you can.

mtreinish:
> Unlikely to be able to do this soon, sorry. But please test interactive usage (completion etc.) with pdbpp if you can.

I just installed pdbpp in my tox venv made the following diff on top of this PR:

```diff
diff --git a/stestr/tests/base.py b/stestr/tests/base.py
index 38bfa83..cae3155 100644
--- a/stestr/tests/base.py
+++ b/stestr/tests/base.py
@@ -18,9 +18,9 @@ class TestCase(testtools.TestCase):
 
     def setUp(self):
         super(TestCase, self).setUp()
-        stdout = self.useFixture(fixtures.StringStream('stdout')).stream
-        self.useFixture(fixtures.MonkeyPatch('sys.stdout', stdout))
-        stderr = self.useFixture(fixtures.StringStream('stderr')).stream
-        self.useFixture(fixtures.MonkeyPatch('sys.stderr', stderr))
-        self.useFixture(fixtures.LoggerFixture(nuke_handlers=False,
-                                               level=None))
+#        stdout = self.useFixture(fixtures.StringStream('stdout')).stream
+#        self.useFixture(fixtures.MonkeyPatch('sys.stdout', stdout))
+#        stderr = self.useFixture(fixtures.StringStream('stderr')).stream
+#        self.useFixture(fixtures.MonkeyPatch('sys.stderr', stderr))
+#        self.useFixture(fixtures.LoggerFixture(nuke_handlers=False,
+#                                               level=None))
diff --git a/stestr/tests/test_config_file.py b/stestr/tests/test_config_file.py
index 31e364f..3ae5398 100644
--- a/stestr/tests/test_config_file.py
+++ b/stestr/tests/test_config_file.py
@@ -57,6 +57,8 @@ class TestTestrConf(base.TestCase):
                                           'fake_test_path')
         # Ensure TestProcessorFixture is created with defaults except for where
         # we specfied and with the correct python.
+        import pdb
+        pdb.set_trace()
         mock_TestProcessorFixture.assert_called_once_with(
             None, command, "--list", "--load-list $IDFILE",
             mock_get_repo_open.return_value, black_regex=None,
```

and ran ``tox -epy37 -- --pdb stestr.tests.test_config_file`` and it looks like the tab completion works fine. It auto tab completes variables fine and edit the history. Although when I did tab completion, I'd get output like:

```
(Pdb++) pl** error when importing ~/.fancycompleterrc.py: [Errno 2] No such file or directory: '/home/mtreinish/.fancycompleterrc.py' **
atform
```
The error between `**` happens whenever I hit tab, but I assume that's just a local setup problem and not because of this PR. Not sure what else I should test, because it's not really a workflow I use so I'm sure that I'm not thoroughly testing the use case.

aspiers:
That's great, thank you so much.  I don't think it needs thorough testing on top of what you already did, just a quick check that basic functionality like (say) stepping through execution, printing the stack trace, setting breakpoints all work as normal.

mtreinish:
@aspiers sorry I missed your reply, yeah the standard debugger functionality like stepping and printing backtraces worked fine too.

masayukig:
The same error(``~/.fancycompleterrc.py`` thing) happened to me. But it worked fine.

PR #279, opened 2019-10-18
Update to recent hacking version

dirkmueller:
This pulls in a newer pycodestyle/flake8 version which is a bit
more strict, so some fallout needs to be fixed.


stestr/scheduler.py
```diff
@@ -46,6 +46,9 @@ def partition_tests(test_ids, concurrency, repository, group_callback,
         :return: A list where each element is a distinct subset of test_ids,
             and the union of all the elements is equal to set(test_ids).
         """
+        def noop():
```
mtreinish: To get this working you'll have to change the function signature to take in a test id. The lambda you're replacing did this with the `_`. To do this here just change this line like so:

```suggestion
        def noop(_):
```

PR #283, opened 2019-11-16
Update test_processor.py

pvinci:
Workaround for change in behavior in python3.6
See: https://bugs.python.org/issue29130


pvinci:
Running tests with stdout/stderr redirection via tox/stestr.  Tests run
under py27 but fail with 120 error under py36.
It may not be your issue in the end, but if we can't run the tests, we
can't fix the underlying issue.

PR #287, opened 2020-01-15
Remove python2 support

### LINKED ISSUE #237: Python 2.7 EoL Planning
With the pending end of life for python 2.7 on January 1st 2020 we need to have a plan for dropping stestr support for python 2.7. Likely at some point in January 2020. Clearly this will warrant a stestr 3.0 release, but we should have the plan written down in the docs or on github somewhere. We also should remove all the six and python 2 compat code from the repo prior to the 3.0 release.

mtreinish:
This commit remove 2.7 support from stestr. This has been advertised since the 2.4.0 release as happening in early 2020 (corresponding to the upstream EoL of python 2.7). It also takes the opportunity to remove all the 2.7 compat code left in the repo which is now no longer necessary.

Fixes #237

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ pbr!=2.1.0,>=2.0.0,!=4.0.0,!=4.0.1,!=4.0.2,!=4.0.3 # Apache-2.0
 cliff>=2.8.0 # Apache-2.0
 python-subunit>=1.3.0 # Apache-2.0/BSD
 fixtures>=3.0.0 # Apache-2.0/BSD
-six>=1.10.0 # MIT
 testtools>=2.2.0 # MIT
 PyYAML>=3.10.0 # MIT
 voluptuous>=0.8.9 # BSD License
--- a/stestr/subunit_runner/program.py
+++ b/stestr/subunit_runner/program.py
@@ -102,11 +102,11 @@ def list_test(test):
     :return: A tuple of test ids that would run and error strings
         describing things that failed to import.
     """
-    unittest_import_strs = set([
+    unittest_import_strs = {
         'unittest2.loader.ModuleImportFailure.',
         'unittest.loader.ModuleImportFailure.',
         'discover.ModuleImportFailure.'
-        ])
+        }
     test_ids = []
     errors = []
     for test in iterate_tests(test):
@@ -178,7 +178,7 @@ class TestProgram(unittest.TestProgram):
                 lines = source.readlines()
             finally:
                 source.close()
-            test_ids = set(line.strip().decode('utf-8') for line in lines)
+            test_ids = {line.strip().decode('utf-8') for line in lines}
             self.test = filter_by_ids(self.test, test_ids)
         # XXX: Local edit (see http://bugs.python.org/issue22860)
         if not self.listtests:
--- a/stestr/subunit_runner/run.py
+++ b/stestr/subunit_runner/run.py
@@ -84,15 +84,9 @@ class SubunitTestRunner(object):
 
 def main():
     runner = SubunitTestRunner
-    if sys.version_info[0] >= 3 and sys.version_info[1] >= 5:
-        program.TestProgram(
-            module=None, argv=sys.argv,
-            testRunner=partial(runner, stdout=sys.stdout))
-    else:
-        from testtools import run as testtools_run
-        testtools_run.TestProgram(module=None, argv=sys.argv,
-                                  testRunner=runner,
-                                  stdout=sys.stdout, exit=False)
+    program.TestProgram(
+        module=None, argv=sys.argv,
+        testRunner=partial(runner, stdout=sys.stdout))
 
 
 if __name__ == '__main__':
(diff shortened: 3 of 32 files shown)


mtreinish:
Yeah, we just need to remove the six usage from the tests first.

masayukig:
thanks, sure. (I just missed the "WIP" tag :-p

mtreinish:
@masayukig I think this is ready to go now

masayukig:
Thanks! I'll review this tomorrow.

stestr/subunit_runner/program.py
```diff
@@ -102,11 +102,11 @@ def list_test(test):
     :return: A tuple of test ids that would run and error strings
         describing things that failed to import.
     """
-    unittest_import_strs = set([
+    unittest_import_strs = {
         'unittest2.loader.ModuleImportFailure.',
```
gmannos: can we remove this now ?

stestr/subunit_runner/run.py
```diff
@@ -84,15 +84,9 @@ def _list(self, test):
...
-            module=None, argv=sys.argv,
-            testRunner=partial(runner, stdout=sys.stdout))
-    else:
-        from testtools import run as testtools_run
-        testtools_run.TestProgram(module=None, argv=sys.argv,
-                                  testRunner=runner,
-                                  stdout=sys.stdout, exit=False)
+    program.TestProgram(
+        module=None, argv=sys.argv,
+        testRunner=partial(runner, stdout=sys.stdout))
```
gmannos: +1, this part was very flaky.

stestr/subunit_runner/program.py
mtreinish: I'm not sure, maybe. I'd have to test a testtools suite (and a direct unittest2 suite) with the stestr subunit runner to see if we can trigger import failures during discovery to see the mangled test id.I'm not sure how important it is though, this is to check for import errors during discovery and treat them as test errors (instead of trying to run them and erroring with an odd importerror). I don't think there is any harm in being slightly conservative and just leaving it in the set, there is no real penalty for having an extra entry in the set even if it never matches.

gmannos: Yeah, no big deal to keep it, thanks.

requirements.txt
```diff
@@ -6,7 +6,6 @@ pbr!=2.1.0,>=2.0.0,!=4.0.0,!=4.0.1,!=4.0.2,!=4.0.3 # Apache-2.0
 cliff>=2.8.0 # Apache-2.0
 python-subunit>=1.3.0 # Apache-2.0/BSD
 fixtures>=3.0.0 # Apache-2.0/BSD
-six>=1.10.0 # MIT
```
masayukig: tools/testr_to_stestr.py still imports six for `six.moves.configparser.ConfigParser()`. How about changing it to `import configparser`, instead?

```
tools/testr_to_stestr.py
18:import six
24:    config = six.moves.configparser.ConfigParser()
```

stestr/tests/test_return_codes.py
```diff
@@ -245,16 +243,16 @@ def test_combine_results(self):
         self.assertRunExit('stestr run passing', 0)
         stdout = self._get_cmd_stdout(
             'stestr last --no-subunit-trace')
-        stdout = six.text_type(stdout[0])
+        stdout = str(stdout[0])
         test_count_split = stdout.split(' ')
         test_count = test_count_split[1]
         test_count = int(test_count)
-        id_regex = re.compile('\(id=(.*?)\)')
+        id_regex = re.compile(r'\(id=(.*?)\)')
```
masayukig: I'm not sure we need these changes(adding `r`) now. But it should work correctly.
We still have a lot of `re.compile` in the rest of the codes, and we can update them later if we need it.

requirements.txt
mtreinish: Ah, good catch. Yeah I didn't do a grep in `tools/`. This should be fixed in: https://github.com/mtreinish/stestr/pull/287/commits/e6ce637a575eb2f58d1df58dcdc93e9b33da8307

stestr/tests/test_return_codes.py
mtreinish: Yeah, I just ran pyupgrade (which might not have been the best idea) which did this conversion to raw strings because we're escaping the `'('`. I'm not sure it hurts to do this, it should only be needed if we are escaping anything in the regex.

gmannos:
I think you missed sphinx requirements cleanup for py2 in test-requirements.txt

mtreinish:
> I think you missed sphinx requirements cleanup for py2 in test-requirements.txt

Yep, thanks for the reminder. I updated it in: https://github.com/mtreinish/stestr/pull/287/commits/82fe384b321179402ab92603d8ba739e3385eb62

I also realized we can and should remove the mock library so I took care of that in the same commit.

stestr/tests/test_return_codes.py
masayukig: ok, sure

PR #289, opened 2020-03-18
Bump python-subunit minimum to 1.4.0

stephenfin:
This version includes an important Python 3 fix [1], without which a Python 3-only stestr doesn't really make sense.

[1] https://github.com/testing-cabal/subunit/pull/40


stephenfin:
\o/ Just need a release now :smile:

PR #297, opened 2020-10-07
Change option '--whitelist-file' to '--include-list'

### LINKED ISSUE #296: Change towards usage of more inclusive and appropriate terms in code and docs
**Is your feature request related to a problem? Please describe.**
This feature looks to rename all uses of "blacklist" and "whitelist" to more accurate and appropriate names.

**Describe the solution you'd like**
Suggestion to replace the below options for `list` and `run` commands as follows:
```
--blacklist-file	>>	--exclude-list
--whitelist-file	>>	--include-list
--black-regex		>>	--exclude-regex
```

As suggested by Matt, we can go ahead and initiate a deprecation notice for the existing options while introducing the alternative options. In a later release, we can remove the deprecated options and make the new options mainline...

arun-n2020:
Added the new option while posting a Deprecation Warning for the old one.
The old option will be supported for now but it is soon to be retired...
So for now, both options are equally usable and will have the same effect.

Tests relying on the old option are kept intact, will be removed only when
the option itself is removed. New tests are added to verify that the same
behaviour is afforded if the new option is used also.

Closes #296

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -171,19 +171,25 @@ used in conjunction with the normal filters then the regex filters passed in as
 an argument regex will be used for the initial test selection, and the
 exclusion regexes from the blacklist file on top of that.
 
-The dual of the blacklist file is the whitelist file which will include any
-tests matching the regexes in the file. You can specify the path to the file
-with ``--whitelist-file``/``-w``, for example::
+The dual of the blacklist file is the inclusion list file which will include
+any tests matching the regexes in the file. You can specify the path to the
+file with ``--include-list``/``-i``, for example::
 
-  $ stestr run --whitelist-file $path_to_file
+  $ stestr run --include-list $path_to_file
 
 The format for the file is more or less identical to the blacklist file::
 
-  # Whitelist File
+  # Inclusion list File
   ^regex1 # Include these tests
   .*regex2 # include those tests
 
 However, instead of excluding the matches it will include them.
+
+.. note::
+    DEPRACATION WARNING:
+    Previously the option ``--whitelist-file``/``-w`` was available for this
+    functionality. While it is still available at this time, it is soon to be
+    replaced by the new (equivalent) option ``--include-list``/``-i``.
 
 It's also worth noting that you can use the test list option to dry run any
 selection arguments you are using. You just need to use ``stestr list``
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -172,7 +172,7 @@ passed in as an argument regex will be used for the initial test selection, and
 the exclusion regexes from the exclusion list file on top of that.
 
 .. note::
-    DEPRACATION WARNING:
+    DEPRECATION WARNING:
     Previously the option ``--blacklist-file``/``-b`` was available for this
     functionality. While it is still available at this time, it is soon to be
     replaced by the new (equivalent) option ``--exclude-list``/``-e``.
@@ -192,7 +192,7 @@ The format for the file is more or less identical to the exclusion list file::
 However, instead of excluding the matches it will include them.
 
 .. note::
-    DEPRACATION WARNING:
+    DEPRECATION WARNING:
     Previously the option ``--whitelist-file``/``-w`` was available for this
     functionality. While it is still available at this time, it is soon to be
     replaced by the new (equivalent) option ``--include-list``/``-i``.
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -138,14 +138,14 @@ filters will be run. For example, if you called ``stestr run foo bar`` this
 will only run the tests that have a regex match with foo **or** a regex match
 with bar.
 
-stestr allows you do to do simple test exclusion via passing a rejection/black
-regexp::
+stestr allows you do to do simple test exclusion via passing a rejection or
+exclusion regexp::
 
-  $ stestr run --black-regex 'slow_tests|bad_tests'
+  $ stestr run --exclusion-regex 'slow_tests|bad_tests'
 
 stestr also allow you to combine these arguments::
 
-  $ stestr run --black-regex 'slow_tests|bad_tests' ui\.interface
+  $ stestr run --exclusion-regex 'slow_tests|bad_tests' ui\.interface
 
 Here first we selected all tests which matches to ``ui\.interface``, then we
 are dropping all test which matches ``slow_tests|bad_tests`` from the final
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -138,7 +138,7 @@ filters will be run. For example, if you called ``stestr run foo bar`` this
 will only run the tests that have a regex match with foo **or** a regex match
 with bar.
 
-stestr allows you do to do simple test exclusion via passing a rejection or
+stestr allows you do to do simple test exclusion via passing a
 exclusion regexp::
 
   $ stestr run --exclusion-regex 'slow_tests|bad_tests'
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -141,11 +141,11 @@ with bar.
 stestr allows you do to do simple test exclusion via passing a
 exclusion regexp::
 
-  $ stestr run --exclusion-regex 'slow_tests|bad_tests'
+  $ stestr run --exclude-regex 'slow_tests|bad_tests'
 
 stestr also allow you to combine these arguments::
 
-  $ stestr run --exclusion-regex 'slow_tests|bad_tests' ui\.interface
+  $ stestr run --exclude-regex 'slow_tests|bad_tests' ui\.interface
 
 Here first we selected all tests which matches to ``ui\.interface``, then we
 are dropping all test which matches ``slow_tests|bad_tests`` from the final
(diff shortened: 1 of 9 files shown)


arun-n2020:
Commits to implement [Issue#296](https://github.com/mtreinish/stestr/issues/296).

doc/source/MANUAL.rst
```diff
@@ -138,58 +138,70 @@ filters will be run. For example, if you called ``stestr run foo bar`` this
 will only run the tests that have a regex match with foo **or** a regex match
 with bar.
 
-stestr allows you do to do simple test exclusion via passing a rejection/black
-regexp::
+stestr allows you do to do simple test exclusion via passing a rejection or
```
mtreinish, line 141: The previous usage of "rejection" only really made sense to explain black-regex, which is just an indication of how bad a name it was:

```suggestion
stestr allows you do to do simple test exclusion via passing a
```

stestr/commands/list.py
```diff
@@ -94,12 +114,19 @@ def list_command(config='.stestr.conf', repo_type='file', repo_url=None,
...
-        on re.search() operation, it will be removed from the final test list.
+    :param str blacklist_file: DEPRECATED: soon to be replaced by the new
+        option exclusion_list_file below.
+    :param str exclusion_list_file: Path to an exclusion list file, this file
+        contains a separate regex exclude on each newline.
+    :param str whitelist_file: DEPRECATED: soon to be replaced by the new
+        option inclusion_list_file below.
+    :param str inclusion_list_file: Path to an inclusion list file, this file
+        contains a separate regex on each newline.
+    :param str black_regex: DEPRECATED: soon to be replaced by the new
```
mtreinish, line 125: We should add a deprecation warning for all these new deprecated arguments. Something like:

```python3
if black_regex is not None:
    warnings.warn("The black_regex argument is deprecated and will be removed in a future "
                  "release. Instead you should use exclude-regex which is functionally equivalent",
                  DeprecationWarning)
```

stestr/commands/list.py
```diff
@@ -37,22 +37,37 @@ def get_parser(self, prog_name):
...
+        parser.add_argument('--include-list', '-i',
+                            default=None, dest='inclusion_list_file',
+                            help='Path to an inclusion list file, this file '
                                  'contains a separate regex on each newline.')
         parser.add_argument('--black-regex', '-B',
                             default=None, dest='black_regex',
+                            help='DEPRECATED: This option will soon be  '
+                            'replaced by --exclusion-regex which is '
+                            'functionally equivalent.')
+        parser.add_argument('--exclusion-regex', '-E',
```
mtreinish, line 62: Instead of `--exclusion-regex` I think `--exclude-regex` is probably a better term. "exclusion" makes more sense in English  but exclude is less typing it also matches `--exclude-list`

stestr/commands/list.py
```diff
@@ -66,15 +81,20 @@ def take_action(self, parsed_args):
...
+                            exclusion_regex=args.exclusion_regex,
                             filters=filters)
 
 
 def list_command(config='.stestr.conf', repo_type='file', repo_url=None,
                  test_path=None, top_dir=None, group_regex=None,
-                 blacklist_file=None, whitelist_file=None, black_regex=None,
-                 filters=None, stdout=sys.stdout):
+                 blacklist_file=None, exclusion_list_file=None,
+                 whitelist_file=None, inclusion_list_file=None,
```
mtreinish, line 95: If we're using `--include-list`  and `--exclude-list` I think the Python API kwargs should match that name and be `include_list` and `exclude_list` just to make it a bit clearer.

stestr/test_processor.py
```diff
@@ -68,10 +68,14 @@ class TestProcessorFixture(fixtures.Fixture):
         to use for the run
     :param int concurrency: How many processes to use. The default (0)
         autodetects your CPU count and uses that.
-    :param path blacklist_file: Path to a blacklist file, this file contains a
-        separate regex exclude on each newline.
-    :param path whitelist_file: Path to a whitelist file, this file contains a
-         separate regex on each newline.
+    :param path blacklist_file: DEPRECATED: soon to be replaced by the new
```
mtreinish, line 71: I think you can remove the deprecated args here, we don't treat `test_processor` as a Stable API. See https://stestr.readthedocs.io/en/latest/api.html#internal-apis

stestr/selection.py
```diff
@@ -83,24 +83,28 @@ def _get_regex_from_whitelist_file(file_path):
...
-                   regexes=None, black_regex=None):
+                   regexes=None, black_regex=None, exclusion_list_file=None,
+                   inclusion_list_file=None, exclusion_regex=None):
     """Filters the discovered test cases
 
     :param list test_ids: The set of test_ids to be filtered
-    :param str blacklist_file: The path to a blacklist file
-    :param str whitelist_file: The path to a whitelist file
+    :param str blacklist_file: DEPRECATED: Replaced by exclusion_list_file
+    :param str whitelist_file: DEPRECATED: Replaced by inclusion_list_file
```
mtreinish, line 99: I think you can remove the deprecated args here, we don't treat `construct_list` as a Stable API. See https://stestr.readthedocs.io/en/latest/api.html#internal-apis

stestr/config_file.py
```diff
@@ -83,11 +85,17 @@ def get_run_command(self, test_ids=None, regexes=None,
...
+        :param str blacklist_file: DEPRECATED: soon to be replaced by the new
+            option exclusion_list_file below.
+        :param str exclusion_list_file: Path to an exclusion list file, this
+            file contains a separate regex exclude on each newline.
+        :param str whitelist_file: DEPRECATED: soon to be replaced by the new
+            option inclusion_list_file below.
+        :param str inclusion_list_file: Path to an inclusion list file, this
+            file contains a separate regex on each newline.
+        :param str black_regex: DEPRECATED: soon to be replaced by the new
+            option exclusion_regex below.
```
mtreinish, line 97: I think you can remove the deprecated args here, we don't treat `TestrConf` as a Stable API. See https://stestr.readthedocs.io/en/latest/api.html#internal-apis

mtreinish (changes requested):
Thanks for getting this started! At a high level this LGTM. I left some comments about naming and raising `DeprecationWarnings` on the list command, but they apply to all the stestr commands not just list. I think this will pretty much be ready once we use consistent naming between all the cli args and the python api and have the commands Python API. There are also a couple of internal APIs you don't have bother with deprecation on. You can refer to https://stestr.readthedocs.io/en/latest/api.html#internal-apis for what we treat as internal private interfaces and what is documented as having a stable public interface. If it's a public interface we have to go through a deprecation first.

The last thing we probably need to test and/or define the behavior of is what happens when you   specify both `--blacklist-file` and `--exclude-list` at the same time with different files? While the deprecated names will only be around for the rest of the 3.x.x release series we don't want to have undefined behavior for however long that lasts.

arun-n2020:
> The last thing we probably need to test and/or define the behavior of is what happens when you specify both `--blacklist-file` and `--exclude-list` at the same time with different files? While the deprecated names will only be around for the rest of the 3.x.x release series we don't want to have undefined behavior for however long that lasts.

The way it is coded right now, the option value for the **new** option has higher precedence, in case both options are used! But if you think it is useful, I could look at the <cite>[`add_mutually_exclusive_group`][0]</cite> method to enforce that only one of the two are used...

[0]: https://docs.python.org/3/library/argparse.html#mutual-exclusion

stestr/selection.py
arun-n2020: Changing "deprecation note" to simple mention of new args...

stestr/config_file.py
arun-n2020: Changing "deprecation note" to simple mention of new args...

stestr/test_processor.py
arun-n2020: Changing "deprecation note" to simple mention of new args...

stestr/commands/list.py
arun-n2020: Agreed :smile:
- fixing this across the files changes and
- updating [Issue#296](https://github.com/mtreinish/stestr/issues/296) text also

stestr/commands/list.py
arun-n2020: :+1:

mtreinish:
> The way it is coded right now, the option value for the **new** option has higher precedence, in case both options are used! But if you think it is useful, I could look at the [`add_mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#mutual-exclusion) method to enforce that only one of the two are used...

No I think this is fine as is, but we should just call it out in the documentation then and also ensure we have a test asserting this behavior just to make sure the edge case behavior is defined.

stestr/commands/list.py
arun-n2020: 👍

arun-n2020:
> > The way it is coded right now, the option value for the **new** option has higher precedence, in case both options are used! But if you think it is useful, I could look at the [`add_mutually_exclusive_group`](https://docs.python.org/3/library/argparse.html#mutual-exclusion) method to enforce that only one of the two are used...
> 
> No I think this is fine as is, but we should just call it out in the documentation then and also ensure we have a test asserting this behavior just to make sure the edge case behavior is defined.

Tests also added now, to validate this precedence, for as long as the deprecation notice holds 👍

mtreinish (approved):
LGTM, thanks for tackling this and keeping on top of the review comments.

arun-n2020:
Thanks Matt☺️👍

PR #298, opened 2020-10-28
Add python 3.9 support

mtreinish:
This commit adds Python 3.9 support to stestr. It adds Python 3.9 as
supported python version in the trove classifiers for the package
metadata and adds CI jobs that test stestr on python 3.9.


masayukig (approved):
Thanks! LGTM

PR #301, opened 2021-02-22
Remove subprocess usage from discovery

mtreinish:
This commit removes the usage of subprocess from the discovery/list test
phase in TestProcessor. Previously to get a list of tests from discovery
we would subprocess out to a new python interpreter calling the subunit
runner with the list command and capture stdout from process and parse
that sububit stream to get the test ids. However, this was not strictly
necessary as we're not doing anything in parallel, so we should be able
to just call the subunit_runner directly inline and use an in memory
buffer instead of stdout. This has the advantage of not interfering
with sys.stdout, in the previous method if anything besides the
subunit_runner was printing to stdout it would cause a parser error
because of the non-subunit data mixed into the stream.

### DIFF
--- a/stestr/test_processor.py
+++ b/stestr/test_processor.py
@@ -10,6 +10,7 @@
 # License for the specific language governing permissions and limitations
 # under the License.
 
+from functools import partial
 import io
 import os
 import re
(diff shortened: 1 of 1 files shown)


mtreinish:
This is still failing some tests, it looks like there is an issue when calling from a function  with the import paths that we'll need to debug first.

PR #302, opened 2021-03-09
Update python syntax to latest

mtreinish:
Since stestr 3.0.0 stestr has only supported Python>=3.5 and we've
already removed a bunch of compatibility code for Python 2 support.
However, there were a lot of places in the stestr code where we were
using older connventions and syntax when there was newer conventions
available that typically are more consise. This commit updates the
syntax to use python3 conventions everywhere. This was done by using the
pyupgrade [1] tool and then fixing places where it didn't work as
expected.

[1] https://pypi.org/project/pyupgrade/

### DIFF
--- a/stestr/bisect_tests.py
+++ b/stestr/bisect_tests.py
@@ -17,12 +17,12 @@ import testtools
 from stestr import output
 
 
-class IsolationAnalyzer(object):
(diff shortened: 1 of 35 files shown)


masayukig (approved):
LGTM, it's simpler :)

PR #304, opened 2021-03-31
Report tests stuck as inprogress as failures

mtreinish:
In some situations (especially when testing compiled extensions) a test
case can trigger a segfault or some other scenario that crashes python.
When this occurs the worker doesn't send back the second subunit event
that reports the test status and that leaves the test case stuck as
inprogress. This becomes tricky to debug as there is no clear message as
to why the test run failed. While we were correctly not returning a 0
return code in this situation there was no error message about why we
didn't. This commit fixes this oversight by adding a check to the end of
subunit-trace so that we report any tests with an inprogress state and
explain that we're failing because they likely crashed python.

### DIFF
--- a/stestr/subunit_trace.py
+++ b/stestr/subunit_trace.py
@@ -269,6 +269,18 @@ def count_tests(key, value):
     return count
 
 
+def get_stuck_in_progress():
+    key = 'status'
+    match = re.compile('^inprogress$')
+    in_progress = []
+    for k, v in RESULTS.items():
+        for item in v:
+            if key in item:
+                if match.search(item[key]):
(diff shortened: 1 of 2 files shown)


mtreinish:
> I'm just curious, how noticed this. Did you face this situation?

Yeah, it's something that was coming up pretty frequently in https://github.com/Qiskit/qiskit-aer which is primarily C++ that gets exposed as a Python extension. A bug in a PR there would often lead to a segfault and the lack of an error message was very confusing to debug. For example: https://github.com/Qiskit/qiskit-aer/runs/2237779093#step:8:1075 where it failed but the summary lists no test failures.

PR #305, opened 2021-04-11
Add support for setting stestr config in a tox.ini file

### LINKED ISSUE #300: Add support for stestr config embedded in tox.ini
While we support having an .steatr.conf and putting that in any path in some use cases if tox is being used it would be nice if we didn't require a separate config file and just put the stestr config in an `[stestr] ` section in the existing tox.ini.

mtreinish:
This commit adds support for setting the stestr config in a tox.ini
file. Using tox to run tests with stestr is a very common pattern and
instead of requiring users create a separate per repo config file we
can just integrate stestr's configuration in the tox.ini config.

Fixes #300


mtreinish:
The one thing I'm wondering if we want to fix here is using tox.ini substitutions: https://tox.readthedocs.io/en/latest/config.html#globally-available-substitutions ? Right now it's fairly naive it just parses the `[stestr]` section in the `tox.ini` as if it were an `stestr.conf` and no substitutions are supported. There are 2 paths forward I can think of, either add a tox mode to the `TestrConf` class that will try to use the global substitutions if set, or just add a note to the manual about the current limitation. I'm leaning more towards the latter, because I don't think we want to import tox's config parser and if we aren't using tox there is a good chance we could diverge from the expected tox behavior over time.

README_ja.rst
```diff
@@ -85,8 +85,15 @@ stestr のインストール後、テスト実行のために使う方法は、
...
-このファイルを作成すれば、stestr を使い始めるためにやるべきことはすべて完了
-です。テストを実行するためには、単に次のように使うだけです::
+あるいは、`tox <https://tox.readthedocs.io/en/latest/>`__
+を使用している場合は、tox.ini ファイルを使用してstestrを構成できます。
+たとえば::
+
+  [stestr]
+  test_path=./project_source_dir/tests
+
+すれば、stestr を使い始めるためにやるべきことはすべて完了です。テストを実行するためには、
```
masayukig, line 95:
```suggestion
と設定すれば、stestr を使い始めるためにやるべきことはすべて完了です。テストを実行するためには、
```
This is kind of a nitpick but I just added "と設定" for "is configured".

masayukig (changes requested):
LGTM, over all. But I have a suggestion about the Japanese translation.

masayukig (approved):
Thanks, LGTM

PR #306, opened 2021-04-11
Add history commands to stestr

### LINKED ISSUE #303: Add history trim option
**Is your feature request related to a problem? Please describe.**

The stestr run history continues to grow over time and can take up multiple gigabytes to store the subunit streams of all the past runs.

**Describe the solution you'd like**
An stestr command to clear out the N oldest runs or to also delete all the subunit history (but keep the timings db). Something like:

```bash
stestr clear all
```

would delete all the subunit history

```bash
stestr clear 10
```

would delete the 10 oldest runs. The exact syntax can change as I think there are a lot of different ways to slice things up and finding something flexible but easy to use might be a bit trickier.

**Describe alternatives you've considered**
You can just do this with a bash script pretty easily since the on-disk structure for the file repository is just integers, but it would be nice to have an integrated command to make this simpler. Also for the sql repository type (assuming we ever finish it) having the command would give a consistent interface between both repository types.

mtreinish:
This commit adds a new command to stestr, history. The history command
is used for interacting with the history of results in the repository.
Prior to this commit you could only interact with the most recent run in
the repository via the last command, but anything older was just taking
up disk space and never used. The history command to start has 3
subcommands, 'list', 'show', and 'remove'. List will generate a table of
all the contents of the repository that will show the run id, the result
of the run, when it started, how long it took, and optionally show any
metadata for the run. Show will show a run in the repository, this is
equivalent to 'stestr last', but takes an optional run id instead of just
unconditionally showing the last run. Remove is used to remove runs from
the repository, this is useful if you don't need the history anymore and
want to clean up some disk space.

Fixes #303
Starts #224

TODO:

- [x] Add tests
- [x] Add user config support for default outputs
- [x] Add support for subunit2sql repositories
- [x] Fix docs builds (looks like it's caused by subcommands on a cliff `Command` breaking the sphinx extension)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,6 @@
 # process, which may cause wedges in the gate later.
 future
 pbr!=2.1.0,>=2.0.0,!=4.0.0,!=4.0.1,!=4.0.2,!=4.0.3 # Apache-2.0
-PrettyTable>=0.7.2 # BSD
 cliff>=2.8.0 # Apache-2.0
 python-subunit>=1.4.0 # Apache-2.0/BSD
 fixtures>=3.0.0 # Apache-2.0/BSD
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,6 +3,7 @@
 # process, which may cause wedges in the gate later.
 future
 pbr!=2.1.0,>=2.0.0,!=4.0.0,!=4.0.1,!=4.0.2,!=4.0.3 # Apache-2.0
+PrettyTable>=0.7.2 # BSD
 cliff>=2.8.0 # Apache-2.0
 python-subunit>=1.4.0 # Apache-2.0/BSD
 fixtures>=3.0.0 # Apache-2.0/BSD
--- a/stestr/commands/history.py
+++ b/stestr/commands/history.py
@@ -1,0 +1,383 @@
+# Licensed under the Apache License, Version 2.0 (the "License"); you may
+# not use this file except in compliance with the License. You may obtain
+# a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations
+# under the License.
+
+"""Interact with the run history in a repository."""
+
+import functools
+import sys
+
+from cliff import command
+import prettytable
+import subunit
+import testtools
+
+from stestr import output
+from stestr.repository import abstract
+from stestr.repository import util
+from stestr import results
+from stestr import subunit_trace
+from stestr import user_config
+
+
+class History(command.Command):
+    """Interact with the run history in a repository."""
+
+    def get_parser(self, prog_name):
+        parser = super().get_parser(prog_name)
+        cmd_parser = parser.add_subparsers()
+        list_parser = cmd_parser.add_parser('list',
+                                            help='list runs in repository')
+        list_parser.set_defaults(func='list')
+        list_parser.add_argument('--show-metadata', '-m', action='store_true',
+                                 help="Show metadata associated with runs in "
+                                      "output table")
+
+        # Configure show subcommand:
+        show_parser = cmd_parser.add_parser('show',
+                                            help='Show the contents of a '
+                                                 'single run in the history')
+        show_parser.add_argument(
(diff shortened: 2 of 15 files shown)


mtreinish:
The CI failures here are blocked on #308 the sql repository tests are failing on windows because something is holding onto a sqlite session too long and windows doesn't let 2 things have an open fd on the same file at once. #308 deprecates the sql repository type and removes the tests from the windows and linux jobs. But aside from that I think this is ready for review.

stestr/commands/history.py
```diff
@@ -0,0 +1,389 @@
...
+# License for the specific language governing permissions and limitations
+# under the License.
+
+"""Interact with the run history in a repository."""
+
+import functools
+import sys
+
+from cliff import command
+import prettytable
```
masayukig, line 19: How about using cliff lister instead of using prettytable directly? That has formatters like csv, value, etc.

```suggestion
from cliff import lister
```

stestr/commands/history.py
```diff
@@ -0,0 +1,389 @@
...
+
+from stestr import output
+from stestr.repository import abstract
+from stestr.repository import util
+from stestr import results
+from stestr import subunit_trace
+from stestr import user_config
+
+
+class HistoryList(command.Command):
```
masayukig, line 31:
```suggestion
class HistoryList(lister.Lister):
```

stestr/commands/history.py
```diff
@@ -0,0 +1,389 @@
...
+    :return return_code: The exit code for the command. 0 for success and > 0
+        for failures.
+    :rtype: int
+    """
+
+    table = prettytable.PrettyTable()
+    if show_metadata:
+        table.field_names = ['Run ID', 'Passed', 'Runtime', 'Date', 'Metadata']
+    else:
+        table.field_names = ['Run ID', 'Passed', 'Runtime', 'Date']
```
masayukig, line 234:
```suggestion
    field_names = ()
    if show_metadata:
        field_names = ('Run ID', 'Passed', 'Runtime', 'Date', 'Metadata')
    else:
        field_names = ('Run ID', 'Passed', 'Runtime', 'Date')
```

stestr/commands/history.py
```diff
@@ -0,0 +1,389 @@
...
+        stream = run.get_subunit_stream()
+        data = _get_run_details(stream, stdout)
+        if show_metadata:
+            table.add_row([run_id, data['passed'], data['runtime'],
+                           data['start'], run.get_metadata()])
+        else:
+            table.add_row([run_id, data['passed'], data['runtime'],
+                           data['start']])
+    stdout.write(table.get_string() + '\n')
+    return 0
```
masayukig, line 257:
```suggestion
    rows = []
    for run_id in run_ids:
        run = repo.get_test_run(run_id)
        stream = run.get_subunit_stream()
        data = _get_run_details(stream, stdout)
        if show_metadata:
            rows.append((run_id, data['passed'], data['runtime'],
                           data['start'], run.get_metadata()))
        else:
            rows.append((run_id, data['passed'], data['runtime'],
                           data['start']))

    return (field_names, rows)
```

masayukig (approved):
LGTM. I think we can approve this PR. But I've noticed an enhancement of this feature but we can push patches if it's needed.

@mtreinish What do you think about using cliff lister?

stestr/commands/history.py
mtreinish: That's a good call, we definitely should just use the cliff lister. This was left over from the first commit on this PR where I was manually building sub-commands via argparse, and using a lister wasn't applicable there because there was only a single command. But since moving to separate cliff commands there's no reason to do this manually anymore.

stestr/repository/file.py
```diff
@@ -110,6 +110,22 @@ def latest_id(self):
             raise KeyError("No tests in repository")
         return result
 
+    def get_run_ids(self):
+        result = []
+        max_id = self._next_stream()
+        if max_id < 0:
+            raise KeyError("No tests in repository")
```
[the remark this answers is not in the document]
mtreinish: Yeah, it's definitely a pathological condition. My concern was more that someone could hand edit the `next-stream` file in the repository dir and set it to -1 or something.

requirements.txt
```diff
@@ -3,6 +3,7 @@
 # process, which may cause wedges in the gate later.
 future
 pbr!=2.1.0,>=2.0.0,!=4.0.0,!=4.0.1,!=4.0.2,!=4.0.3 # Apache-2.0
+PrettyTable>=0.7.2 # BSD
```
mtreinish, line 6:
```suggestion
```

PR #308, opened 2021-04-12
Deprecate the sql repository type

### LINKED ISSUE #307: Deprecate subunit2sql repository?
When we first started stestr the long term plan was to replace the file repository with a new repository built around subunit2sql which provides a much richer api. But in the years since then subunit2sql hasn't been super active and there are still some api gaps that prevent it from being a complete implementation. So we've had the sql repository type as an experimental optional repository backend. This is an undesirable state as we end up with a dual maintenance burden and any change to the interface we have to duplicate for both file and sql backends. We also have to maintain an optional requirements list and extra options for switching between repository types. At this point it seems unlikely that we'll ever end up finishing the sql repository backend, so I'm wondering if we should just call it and deprecate the backend for future removal.

mtreinish:
This commit deprecates the sql repository type for future removal in the
next major version release (4.0.0). Included in this deprecation is the
`--repo-type`argument which only makes sense with >1 repository type. The
sql repository type depends on the subunit2sql project which is not
really active anymore. It never managed to reach feature parity with the
default file repository type and has been listed as experimental it's
whole existence. At this point it's realistically never going to replace
the file repository which was the original long term plan. One note is
that we both emit a `DeprecationWarning` and print a warning to stderr
when the sql repository type is set. This is because in my local testing
the `DeprecationWarning` was not actually show to the user, but I left it
in place because it is the standard python mechanism for these things.

Additionally, since we're deprecating the repository type to save some
CI time and complexity I removed the sql repository tests from the
default test run. We still run it in the macOS CI jobs to ensure we
don't regress during the deprecation but we don't need to spend time
running it in every run.

Fixes #307

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -35,7 +35,7 @@ jobs:
       - name: Install and run tests macOS
         run: |
           tox -epy --notest
-          .tox/py/bin/pip install gnureadline
+          .tox/py/bin/pip install gnureadline subunit2sql
           tox -epy
         if: runner.os == 'macOS'
   lint:
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -621,12 +621,13 @@ contains the following files:
 SQL
 '''
 This is an experimental repository backend, that is based on the `subunit2sql`_
-library. It's currently still under development and should be considered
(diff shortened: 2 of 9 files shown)


masayukig (approved):
LGTM

masayukig (approved):
LGTM.

PR #312, opened 2021-04-14
Update package metadata field names

mtreinish:
In recent version of setuptools installing stestr has started raising
deprecation warnings about some of the metadata fields we were setting.
This commit updates the setup.cfg to use the names that are not
deprecated to fix the warnings and avoid having issues with a future
release of setuptools when the fields are removed.


masayukig (approved):
LGTM

PR #317, opened 2022-01-05
Remove deprecated command line flags

mtreinish:
In preparation for the stestr 4.0.0 release this commit removes the
deprecated command line options. This includes '--whitelist-file',
'--blacklist-file', '--black-regex' and '--partial'. The selection
options have been deprecated for > 1 year and the '--partial' option has
been deprecated > 4 years and hasn't had any effect in that time. The
4.0.0 release will be primarily to make these deprecation removals
(along with the sql repository type which will be in a separate commit).

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -190,12 +190,6 @@ list file is used in conjunction with the normal filters then the regex filters
 passed in as an argument regex will be used for the initial test selection, and
 the exclusion regexes from the exclusion list file on top of that.
 
-.. note::
-    DEPRECATION WARNING:
-    Previously the option ``--blacklist-file``/``-b`` was available for this
-    functionality. While it is still available at this time, it is soon to be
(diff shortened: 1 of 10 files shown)


mtreinish:
Oops, I forgot to update the docs too there are still references in the manual on the deprecated names

mtreinish:
> Oops, I forgot to update the docs too there are still references in the manual on the deprecated names

Ok, fixed this in the most recent push

masayukig (approved):
LGTM

JFYI, the unit tests work on python3.10 in my environment, too.

mtreinish:
> JFYI, the unit tests work on python3.10 in my environment, too.

Cool, yeah we should push up a pr before we prepare 4.0.0 to add 3.10 to ci and add it to the trove classifiers in the setup.cfg. We probably should drop 3.5 at the same time. Do you want to push a pr doing this? Otherwise I can do it later this week.

masayukig:
Thanks! I pushed that #319 .

PR #318, opened 2022-01-05
Remove deprecated sql repository type and repo_type options

mtreinish:
In preparation for the 4.0.0 release this commit removes the deprecated
sql repository type (which is based on subunit2sql) and the repo_type
cli flag and api parameters. Both of these have been deprecated for > 1
year. The sql repository type never reached feature pairity with the
default file repository, primarily because subunit2sql development
stalled (which is mostly my fault for moving on to new projects). As
it was never really a viable user option because of this and
maintaining the dual repository type path was adding additional
development overhead for additional features we were hoping to
implement it was deprecated in preparation for removal in the 4.0.0
release.


stestr/cli.py
```diff
@@ -68,14 +68,14 @@ def _set_common_opts(self, parser):
                                  "command. If one isn't specified then "
                                  ".stestr.conf in the directory that a command"
                                  " is running from is used")
-        parser.add_argument('--repo-type', '-r', dest='repo_type',
-                            choices=['file', 'sql'], default='file',
-                            help="DEPRECATED: Select the repo backend to use")
         parser.add_argument('--repo-url', '-u', dest='repo_url',
                             default=None,
                             help="Set the repo url to use. An acceptable value"
```
masayukig, line 73: I feel it's good to remove this sentence to avoid confusion - "An acceptable value for this depends on the repository type used.". New users don't know about the ``--repo-type`` option any more.

masayukig (changes requested):
LGTM, overall. But I put a couple of comments about the doc and help. I think it's better to be updated.

stestr/cli.py
mtreinish: Fixed in: https://github.com/mtreinish/stestr/pull/318/commits/46e4c46be25c0bbc6668ced292ac45c09fef7b45

mtreinish:
Sorry for the delay, I forgot to circle back and update this based on your comments. It should be fixed now.

stestr/repository/util.py
```diff
@@ -12,15 +12,10 @@
...
-import sys
-import warnings
 
 
 def _get_default_repo_url(repo_type):
-    if repo_type == 'sql':
-        repo_file = os.path.join(os.getcwd(), '.stestr.sqlite')
-        repo_url = 'sqlite:///' + repo_file
-    elif repo_type == 'file':
+    if repo_type == 'file':
```
masayukig, line 18: I'm sorry for missing this at the first review. But I think we probably can remove this `_get_default_repo_url()` method and `repo_type` parameters in `get_repo_open()` and `get_repo_initialise()` because this method only accepts `repo_type=='file'` now, and we only use `repo_type=memory` internally.

Do we still need these `repo_type` for backward compatibility or something?

masayukig (commented):
I think we can clean up the `repo_type` later after this patch getting merged. But I'd like to know your opinion just in case.

stestr/repository/util.py
mtreinish: I think you're right `_get_default_repo_url()` is definitely not needed anymore and the `repo_type` argument on the other functions isn't really used anymore.

I didn't remove the `repo_type` argument from `get_repo_open()` or `get_repo_initialise()` primarily because I was concerned about backwards compatibility. I don't remember if we explicitly made these functions part of the public python api or not, but since we had the deprecation warning for `repo_type='sql'` I figured it was better to be safe and not change the function signature in this PR.

mtreinish: I think maybe before we release 4.0.0 we should deprecate the repo type argument. If you have time and want to open a PR for it that would be great.

masayukig: Thanks, yeah, I'll try that.

PR #319, opened 2022-01-06
Add Python 3.10 support and drop 3.5

masayukig:
This commit adds Python 3.10 support and drop 3.5 at the same time. We
should support the latest release version as possible. And, we can drop
the 3.5 because the version is already EOL for more than 1.5 years[1].

[1] https://www.python.org/downloads/release/python-3510/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9, 3.10]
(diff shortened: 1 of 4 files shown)


masayukig:
Hmm, why are these tests required even if I removed the version from main.yml?

> tests-python3.5-macOS-latest Expected — Waiting for status to be reported
Required
tests-python3.5-ubuntu-latest Expected — Waiting for status to be reported
Required
tests-python3.5-windows-latest Expected — Waiting for status to be reported
Required

masayukig:
@mtreinish If you notice something about this, please feel free to update this PR if you want.

mtreinish (approved):
LGTM, thanks for handling this. Those lines were coming from the branch protection rules which were requiring the 3.5 jobs to pass before we could merge this. I updated the rules to no longer include the 3.5 jobs in the required list.

PR #325, opened 2022-04-25
Add coverxml task and cover job

masayukig:
This commit adds coverxml task to tox.ini and cover job to main.yml to
enable coverage tracking. We already have it in travis.yml but it seems
it hasn't worked recently. And, it's a good timing to migrate to GitHub
Actions if it works well because we can remove travis.yml just only for
tracking the coverage.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,10 +65,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Set up Python 3.6
+      - name: Set up Python 3.10
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6
+          python-version: 3.10
       - name: Pip cache
         uses: actions/cache@v2
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -68,7 +68,7 @@ jobs:
       - name: Set up Python 3.10
         uses: actions/setup-python@v2
         with:
-          python-version: 3.10
+          python-version: "3.10"
       - name: Pip cache
         uses: actions/cache@v2
         with:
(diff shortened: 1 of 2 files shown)


masayukig:
@mtreinish It seems codecov works well. What do you think? I don't stick on Codecov that much but looking at [the marketplace](https://github.com/marketplace?category=code-quality&query=coverage+sort%3Apopularity-desc&type=actions&verification=), it's the most popular one.

.github/workflows/main.yml
```diff
@@ -60,6 +60,36 @@ jobs:
...
         run: tox -epep8
+  cover:
+    name: cover
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.6
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.6
```
mtreinish, line 71: I don't think we want to use python 3.6 since it's EoL. Maybe lets use the latest, 3.10 instead.

```suggestion
      - name: Set up Python 3.10
        uses: actions/setup-python@v2
        with:
          python-version: 3.10
```

mtreinish (commented):
LGTM, we were using coveralls before but you're right we haven't been actually running it in a while: https://coveralls.io/github/mtreinish/stestr. I'm fine using codecov instead the tools are equivalent to me so I don't have a preference either way.

The only issue I have is the python version we use in the new job if we update that I think this is good to go.

masayukig:
Thanks for your comment and suggestion.
I agree with you.
I'll send another PR to remove `.travis` directory and `.travis.yml` which aren't used anymore after this PR gets merged.
Or I can include it to this PR if you prefer it.

mtreinish (approved):

PR #328, opened 2022-04-28
Remove CallWhenProcFinishes class in utils.py

masayukig:
This commit removes CallWhenProcFinishes class in utils.py, its usage
was removed in the commit[1].

 [1] 182c6acda970088ea36d0252ce2fd08844b3b02d


mtreinish (approved):
LGTM, I don't really remember what this was used for or why exactly I removed its usage ~5 years ago. But we can definitely remove this if it's not being used anymore.

PR #331, opened 2022-08-17
Fix #330 - Preserve load-list order

### LINKED ISSUE #330: Test are run out of order when using `--load-list`
**Issue description**

When using `--load-list` to run tests, the tests run in a different order than the one provided in the file, even when the
`--random` argument is not present.

This worked with os-testr and I think it even worked with stestr at some point before sets were used to remove duplicates.

**Expected behavior and actual behavior**

The expected behavior is that the order of tests in the file is preserved unless the `--random` argument is provided.

**Steps to reproduce the problem**

We create file `testfile` with contents:
```
stestr.tests.test_return_codes.TestReturnCodes.test_history_show_passing
stestr.tests.test_return_codes.TestReturnCodesToxIni.test_serial_fails_from_func
stestr.tests.test_load.TestLoadCommand.test_empty_with_pretty_out
```
Now we run the 2 tests with:

```
$ tox -epy39 -- --serial --load-list testfile
```

And it will most likely run in a different order.

Akrog:
When using `--load-list` to run tests the tests were always run in a different order than the one provided in the file, even when the `--random` argument was not present:

```
  $ stestr run --serial --load-list failing
```

This patch preserves the order of the tests from the file when `--load-list` is used, and it will only randomize it when `--random` is provided:

```
  $ stestr run --serial --random --load-list failing
```

### DIFF
--- a/stestr/subunit_runner/program.py
+++ b/stestr/subunit_runner/program.py
@@ -22,7 +22,7 @@ import extras
 
 
 def filter_by_ids(suite_or_case, test_ids):
-    """Remove tests from suite_or_case where their id is not in test_ids.
+    """Return a test suite with the intersection of suite_or_case and test_ids.
 
     :param suite_or_case: A test suite or test case.
     :param test_ids: Something that supports the __contains__ protocol.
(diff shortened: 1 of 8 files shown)


stestr/subunit_runner/program.py
```diff
@@ -58,23 +58,31 @@ def filter_by_ids(suite_or_case, test_ids):
...
+            _filter_tests(res, id_map)
+        # TestCase objects.
+        elif extras.safe_hasattr(to_filter, 'id'):
+            test_id = to_filter.id()
+            if test_id in test_ids:
+                id_map[test_id] = to_filter
+        # Standard TestSuites or derived classes.
+        elif isinstance(to_filter, unittest.TestSuite):
+            for item in to_filter:
+                _filter_tests(item, id_map)
```
mtreinish, line 74: I'm a little uneasy about the recursion here. I feel like it probably would have been a bit cleaner to just do it iteratively. But it's not a big deal this should be fine. If it becomes an issue we can always refactor it later

mtreinish (approved):
Overall this looks great to me, thanks for tackling this.

PR #332, opened 2022-09-16
Apply black autoformatter to code base

jogo:
https://black.readthedocs.io/en/stable/index.html

Black is quickly becoming the standard python formatter, which makes it easier to write uniformly formatted code without having to manually fix lots of issues by hand.


tox.ini
```diff
@@ -18,6 +18,7 @@ commands =
 sitepackages = False
 commands =
   flake8 {posargs}
+  black --check {posargs} .
```
mtreinish, line 22: Can you add a `black` job here too so people can just run `tox -eblack` and get it to take care of auto formatting. Something like:
```ini
[testenv:black]
envdir = .tox/pep8
commands = black {posargs} stestr tools docs setup.py
```

tox.ini
```diff
@@ -18,6 +18,7 @@ commands =
...
```
mtreinish, line 21: I would change this to be:

```suggestion
  black --check {posargs} stestr tools docs setup.py
```

mostly because with `.` it might pick up on venvs from tox or other random stuff in the repo which could take some time if it contains external packages that don't use black. I think having it cover explicitly the paths in the repo which contain python code makes the most sense.

mtreinish (commented):
Overall this LGTM, I've been using black on other projects for some time, it's fast and makes the formatting easy to work with. While I'm not personally a fan with some of the decisions it makes, not having to worry about code formatting is worth that tradeoff. I just have a couple of small comments in the `tox.ini` but other than that this should be good to merge I think

tox.ini
jogo: done

tox.ini
jogo: done

mtreinish (approved):
LGTM, thanks for doing this

PR #333, opened 2022-09-16
Reuse subunit_trace output/validation logic for load/run

jogo:
Previously `stestr run` and subunit_trace used slightly different output and validation logic. run wouldn't catch the case where a test exited without a status, while code using `last` and `history` would since they used subunit_trace.

Refactor the subunit_trace output and validation logic and use it for `load`/`run`.

Converge on the subunit_trace formatted output, in addition to the summary also show if something goes wrong.

The primary motivation for this change is to better detect when a subprocess is killed and not all tests are run as expected (segfault, OOM etc.).

### DIFF
--- a/stestr/subunit_trace.py
+++ b/stestr/subunit_trace.py
@@ -265,11 +265,7 @@ def show_outcome(
 
 
 def print_fails(stream):
-    """Print summary failure report.
-
-    Currently unused, however there remains debate on inline vs. at end
-    reporting, so leave the utility function for later use.
-    """
+    """Print summary failure report."""
     if not FAILS:
         return
     stream.write("\n==============================\n")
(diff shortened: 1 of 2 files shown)


stestr/subunit_trace.py
```diff
@@ -470,24 +466,29 @@ def trace(
...
-        return 1
     if post_fails:
         print_fails(stdout)
     if not no_summary:
         print_summary(stdout, elapsed_time)
 
     # NOTE(mtreinish): Ideally this should live in testtools streamSummary
     # this is just in place until the behavior lands there (if it ever does)
-    if count_tests("status", "^success$") == 0:
+    if count_tests("status", "^xfail$") + count_tests("status", "^success$") == 0:
```
jogo, line 491: figured this is easier to read than join the two regex

mtreinish: yeah, that's fine I think the regex would just be `^xfail|success$` or something like that, but this is definitely more legible.

mtreinish (approved):
LGTM, thanks for the fix

PR #341, opened 2023-01-28
Add support for storing stestr configs in pyproject.toml

### LINKED ISSUE #340: pyproject.toml support
Everybody loves `pyproject.toml` for storing all of their project dependencies, metadata, and test tooling configs.  `stestr` should support it.

Ideally, it would simply be a `[tool.stestr]` section, similar to how `tox` support works (but TOML instead of `.ini`, obviously)

0xDEC0DE:
- Lightly refactor `config_file.TestrConf` to improve code reuse, and make it easier to invoke and use
- Add/update tests where needed
- Update documentation

Closes #340

### DIFF
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -84,6 +84,8 @@ CLI. This way you can run stestr directly without having to write a config file
 and manually specify the test_path like above with the ``--test-path``/``-t``
 CLI argument.
 
+.. _tox:
+
 Tox
 '''
 
@@ -102,6 +104,22 @@ specified and pointing to an existing file or the default location
 ``.stestr.conf`` file is present then any configuration in the ``tox.ini`` will
 be ignored. Configuration embedded in a ``tox.ini`` will only be used if other
 configuration files are not present.
+
+pyproject.toml
+''''''''''''''
+
+Similarly, if your project is using ``pyproject.toml``, you may forego the
+config file, and instead create a ``[tool.stestr]`` section with the desired
+configuration options.  For example::
+
+  [tool.stestr]
+  test_path = "./project/tests"
+  top_dir = "./"
+  group_regex = "([^\.]*\.)*"
+
+The same caveats apply as the :ref:`tox` with regards to CLI arguments.  Also
+of note is that files specified with ``--config-file``/``-c`` may be ``.ini``
+or TOML format.
 
 Running tests
 -------------
--- a/doc/source/MANUAL.rst
+++ b/doc/source/MANUAL.rst
@@ -92,18 +92,19 @@ Tox
 If you are also using `tox <https://tox.readthedocs.io/en/latest/>`__ with your
 project then it is not necessary to create separate stestr config file, instead
 you can embed the necessary configuration in the existing ``tox.ini`` file with
-an ``stestr`` section. For example a full configuration section would be::
+an ``[stestr]`` section. For example a full configuration section would be::
 
   [stestr]
   test_path=./project/tests
   top_dir=./
   group_regex=([^\.]*\.)*
 
+Any configuration directives outside the ``[stestr]`` section will be ignored.
 It's important to note that if either the ``--config``/``-c`` CLI argument is
-specified and pointing to an existing file or the default location
-``.stestr.conf`` file is present then any configuration in the ``tox.ini`` will
-be ignored. Configuration embedded in a ``tox.ini`` will only be used if other
-configuration files are not present.
+specified, or the default location ``.stestr.conf`` file is present
+then any configuration in the ``tox.ini`` will be ignored. Configuration
+embedded in a ``tox.ini`` will only be used if other configuration
+files are not present.
 
 pyproject.toml
 ''''''''''''''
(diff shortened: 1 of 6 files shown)


doc/source/MANUAL.rst
```diff
@@ -103,6 +105,22 @@ specified and pointing to an existing file or the default location
 be ignored. Configuration embedded in a ``tox.ini`` will only be used if other
 configuration files are not present.
 
+pyproject.toml
+''''''''''''''
+
+Similarly, if your project is using ``pyproject.toml``, you may forego the
+config file, and instead create a ``[tool.stestr]`` section with the desired
+configuration options.  For example::
```
mtreinish, line 113: I think it'll be good to document the priorty order between the `.stestr.conf`, `tox.ini`, and `pyproject.toml`. While I think it's unlikely in practice for someone to configure multiple in a repository being explicit here about how this will be loaded under the covers and making that semi-official will likely prevent issues down the road in case someone does.

doc/source/MANUAL.rst
```diff
@@ -103,6 +105,22 @@ specified and pointing to an existing file or the default location
...
+
+  [tool.stestr]
+  test_path = "./project/tests"
+  top_dir = "./"
+  group_regex = "([^\.]*\.)*"
+
+The same caveats apply as the :ref:`tox` with regards to CLI arguments.  Also
+of note is that files specified with ``--config-file``/``-c`` may be ``.ini``
+or TOML format.
```
mtreinish, line 122: Personally, I'd probably prefer to restrict the toml support to just the pyproject.toml file and leave the `-c` argument only for `ini` format. But I can see how people might prefer `toml` to `ini` and it's not like this real hurts anything. But we should be a bit more explicit here, and say basically this only works if the file extension is `.toml` and the section has to be the same as the pyproject.toml example above `[tool.stestr]` just to make it clear how the toml format support for arbitrary config files works.

mtreinish (commented):
Overall this LGTM, thanks for doing this! I totally agree adding support for configuring stestr via the pyproject.toml makes a ton of sense. Sorry this took me a while to review in depth. I just left a few inline comments mostly about the documentation and testing for the implicit ordering around the 3 default configuration locations. Other than that I think this looks good. I didn't see any issues in the code actually doing the loading.

requirements.txt
```diff
@@ -8,3 +8,4 @@ fixtures>=3.0.0 # Apache-2.0/BSD
 testtools>=2.2.0 # MIT
 PyYAML>=3.10.0 # MIT
 voluptuous>=0.8.9 # BSD License
+tomlkit>=0.11.6 # MIT
```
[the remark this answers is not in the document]
0xDEC0DE: I honestly have no memory of making this choice; from what I can tell looking around at things, I appear to have stolen my implementation from either [Hatch](https://hatch.pypa.io/latest/) or [Poetry](https://python-poetry.org), both of which pull in tomlkit.

Although after-the-fact, I am somewhat partial to the UI of the TOML parser being one line long, and providing a simple dict interface.  But similarly, I am not married to it, and it seems to provide way, way, WAY more functionality than what is needed, so if using tomli makes things simpler overall, I can switch.

stestr/config_file.py
```diff
@@ -27,17 +28,65 @@ class TestrConf:
...
+        - tox.ini with a valid [stestr] section
+
+        :param str config: The pathname of the config file to use
+        """
+        if os.path.isfile(config):
+            return cls(config)
+        try:
+            return cls("pyproject.toml")
+        except (FileNotFoundError, KeyError):
+            return cls("tox.ini", section="stestr")
```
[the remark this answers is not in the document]
0xDEC0DE: That's crazy; I could have sworn I did this, especially since I went through the trouble of making the implicit load order explicit here.  I wonder where it went?  😆

doc/source/MANUAL.rst
0xDEC0DE: Can do.

doc/source/MANUAL.rst
0xDEC0DE: That's fair; I'll see what I can come up with.

requirements.txt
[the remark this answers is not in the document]
mtreinish: Nah, this is fine. I was mostly just curious as I don't really use toml very often and don't have a strong opinion on parser libraries for it so wanted to know if there was a reason to prefer one over the other. Tomlkit seems like a good choice

0xDEC0DE:
Squash+merge is highly recommended for this.

doc/source/MANUAL.rst
```diff
@@ -84,24 +84,57 @@ CLI. This way you can run stestr directly without having to write a config file
...
+The order in which configuration files are read is as follows:
+
+* Any file specified with the ``--config``/``-c`` CLI argument
+* The ``.stestr.conf`` file
+* The ``[tool.stestr]`` section in a ``pyproject.toml`` file
+* The ``[stestr]`` section in a ``tox.ini`` file
+
+Also of note is that files specified with ``--config-file``/``-c``
+may be either ``.ini`` or TOML format.  If providing configs in
+TOML format, they **must** be located in a ``[tool.stestr]`` section.
```
0xDEC0DE, line 135: Ah, whoops, the filename also needs to have a `.toml` extension if providing TOML.  Update incoming...

mtreinish (approved):
This LGTM, thanks for updating this.

Issue #351, opened 2023-09-20
Cannot use pyproject.toml configs on the first run


0xDEC0DE:
There is a small bootstrapping problem with the `pyproject.toml` support.

# Steps to reproduce

* Install stestr 4.1.0
* Add configs to `pyproject.toml`
* `rm -rf .stestr .stestr.conf`
* `stestr run`

# Expected result

It runs the defined tests

# Actual behavior

```
No config file found and --test-path not specified. Either create or specify a .stestr.conf or use --test-path 
```

# Workaround

Providing a `--test-path` switch for the first run will allow the `pyproject.toml` to be used for all runs thereafter.

It is also possible to "fake it" and create a repo directory by hand:
```
mkdir .stestr
echo 0 > .stestr/next-stream
echo 1 > .stestr/format
```

# Miscellany

This appears to be due to some extra bootstrapping code in the `run` workflow that was overlooked, and not under test.

PR #352, opened 2023-09-20
Bugfix: make repo initialization respect pyproject.toml

0xDEC0DE:
Make the repo initialization step look for all supported config formats before giving up.

Fixes: Issue #351

### DIFF
--- a/stestr/commands/run.py
+++ b/stestr/commands/run.py
@@ -12,7 +12,6 @@
 
 """Run a projects tests and load them into stestr."""
 
-import configparser
 import errno
 import functools
 import io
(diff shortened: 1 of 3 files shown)


0xDEC0DE:
I have no earthly idea what codecov wants me to do here.  I have increased coverage, but it's still mad at me.

mtreinish (approved):
LGTM, thanks for fixing this. Don't worry about codecov I think it's just confused because the measured code just had more removals than additions. I thought we had the checks with red x's disabled tbh, we try to treat coverage as informative and not a gating check. I'll check the settings for it.

PR #353, opened 2023-10-03
Test on Python 3.12

cclauss:
https://www.python.org/downloads/release/python-3120/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,16 +10,16 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9, "3.10", "3.11"]
+        python-version: [3.7, 3.8, 3.9, "3.10", "3.11", "3.12"]
         os: ["macOS-latest", "ubuntu-latest", "windows-latest"]
     steps:
-      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


mtreinish:
I think the failures in tests here are because testtools isn't compatible with Python 3.12. This PR: https://github.com/testing-cabal/testtools/pull/353 fixes it on main, but it hasn't been released yet.

jcapiitao:
The patch can be rebased as testtools is now availalble for Python3.12 since 2.7.0 [1]

[1] https://pypi.org/project/testtools/2.7.0/

mtreinish (approved):

jcapiitao:
Thank you guys, do you plan to issue a new release supporting py3.12 soon ? No pressure, just wondering

mtreinish:
We can push out a quick bugfix release if needed, but there weren't any code changes associated with this PR. So the bugfix would just be package metadata. This PR was just adding the 3.12 testing to CI, I believe the 4.1.0 release should be 3.12 compatible. We were blocked on testing in CI before because our upstream dependencies weren't 3.12 compatible in a release yet. But now that's been solved installing 4.1.0 should work. If it doesn't definitely open an issue and we can quickly push out 4.1.1

jcapiitao:
Great, thank you @mtreinish  :+1:

PR #359, opened 2024-01-29
Remove Python 3.7 support

kajinamit:
Python 3.7 already reached its EOL on 2023-06-27[1].

[1] https://devguide.python.org/versions/


mtreinish (approved):
LGTM, we tend to support a python version past the eol date a bit so people can still test code as they migrate to a newer version. But I agree it's been long enough now. Thanks for doing this.

PR #373, opened 2025-04-20
Modernize Python packaging

mtreinish:
Recently we attempted to release stestr 4.2.0 but were blocked because of how PBR was embedding the README file in the package metadata for the sdist was triggering a failure in the metadata checks before the package upload. After fiddling with the formatting this came down to how PBR was indenting the files for upload. Since stestr was originally created the python packaging ecosystem has made a lot of progress and the advantage of PBR (package metadata as static config) isn't really unique anymore and doing this directly via official PyPA maintained tools is now an option. This commit updates the python packaging to move away from using PBR and switches to use flit instead. It removes the setup.py and setup.cfg files and instead relies on the pyproject.toml file for all the package metadata. The only downside with this approach is PBR would hook into git to determine the version number automatically, but flit doesn't have this capability so now the version number must be set manually. Given the low level of activity on stestr development now this is probably fine, but it's something to keep in mind.


masayukig (approved):
thanks for updating! lgtm

PR #376, opened 2025-06-10
Remove support for Python 3.8

kajinamit:
Python 3.8 already reached its EOL on 2024-10-07 [1].

[1] https://devguide.python.org/versions/


mtreinish (approved):
LGTM thanks for doing this

PR #383, opened 2026-05-10
Bump github actions

kajinamit:
Some of these are deprecated because they are dependent on old nodejs version.


mtreinish (approved):
LGTM, thanks for doing this. This is also a good reminder that we probably should setup dependabot to automate this in the future.

